PR #20, opened 2015-03-18
Updates to RabbitMQQueue to use exchange_name config value

shakie:
During some recent Laravel 5 development to replace an existing php queue publisher and consumer, I ran into some issues with the use of the exchange_name config value. This pull requests contains the changes I made in order to get the queue driver working as documented. I have made sure that the fallback behaviour is bc with the current implementation.

I have also allowed for the creation of a routing key config on a per queue basis as opposed to forcing the name of the queue. AMQPConnection has also been updated to AMQPStreamConnection as the former is now deprecated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /vendor
 composer.lock
+/nbproject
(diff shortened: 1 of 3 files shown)


vladimir-yuldashev:
Do we really need to have an opportunity to set own `exchange_name` and `routing_key`?

shakie:
It was certainly a use case for me recently. The documentation also alludes
to the fact that it is possible to specify an exchange but that does not
work currently.

vladimir-yuldashev:
@shakie Just tested your pull request. ``Queue::push`` works, but ``Queue::later`` does not. Please, test ``Queue::later`` and let me know if it works on your machine.

shakie:
Of course, no problem. I will keep you updated.

shakie:
@vladimir-yuldashev Do you still need me to take a look at the issue with `Queue::later`? I've been very busy over the last 2 months and haven' had a chance yet.

vladimir-yuldashev:
Surely

PR #22, opened 2015-03-31
Adding travis-ci for unit test

Grummfy:
For now there is no unit test, but I assume that there will be some later, so here is the travis file.


vladimir-yuldashev:
No need for now.

PR #40, opened 2015-10-23
Expose message properties

SteffenBach:
This provides access for producers and consumers to the message metadata such as `correlation_id`, or as it was relevant to my use case, the `application_headers`.

### DIFF
--- a/src/FintechFab/LaravelQueueRabbitMQ/Queue/Jobs/RabbitMQJob.php
+++ b/src/FintechFab/LaravelQueueRabbitMQ/Queue/Jobs/RabbitMQJob.php
@@ -123,4 +123,25 @@ class RabbitMQJob extends Job implements JobContract
 		return $this->message->get('correlation_id');
 	}
 
+	/**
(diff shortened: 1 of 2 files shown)


SteffenBach:
@vladimir-yuldashev I see that you closed this PR without merging it or giving a reason for rejecting it. 

Do you have some objection to the patch itself? If not, then why was it closed?

vladimir-yuldashev:
`$properties = $properties + $options['properties'];`

This line can give strange result. Better use `array_merge`. Also, `content-type` seems to be correct, but you ask to change to `content_type`. `getProperties` and `getProperty` are not Queue API methods. So I prefer to stick to the API.

SteffenBach:
@vladimir-yuldashev 

I realize that `getProperties` and `getProperty` are outside what is specified by the Laravel API, but so is `getJobId` :smirk: .
I can see, that I don't need them merged upstream if I just extend `RabbitMQJob` in my own glue code.

Will you accept a new PR with just the sending of additional properties if it uses `array_merge` ?

vladimir-yuldashev:
@SteffenBach yes. Use branch 5.2 to start from.

PR #56, opened 2016-04-10
update bugfix delay rabbitmq support lavarel delay function

liuchenrang:
this lib very usefully  , but i found a bug, lavarel rabbit delay cant work! wish u can merge!


vladimir-yuldashev:
Merged and in release 5.2. Thanks!

PR #58, opened 2016-05-16
Toolstation

nicksnellockts:
This updates to use the new format for a job when incrementing the number of requests. It also creates a delayed queue in the same exchange as the master queue.


nicksnellockts:
closing to address merge comnflicts

PR #65, opened 2016-06-23
Fix for attempt counter

mintopia:
The attempt counter is now being persisted as part of the job data, but is not being retrieved correctly from the job. The end result is that the number of attempts is always zero.


vladimir-yuldashev:
On which version of Laravel did you test your fixes?

mintopia:
5.2.39

vladimir-yuldashev:
Okay. Already in 5.2 release. Thanks!

Issue #66, opened 2016-07-06
Undefined index: command


rdrenth:
When I use [Queued Event Listeners](https://laravel.com/docs/5.2/events#queued-event-listeners) I get the error "Undefined index: command" when the handle() function of that event listener fails / throws an exception. 

The method `release()` in the class `VladimirYuldashev\LaravelQueueRabbitMQ\Queue\Jobs\RabbitMQJob` expect there to be `$body['data']['command']`, but when using Queued Event Listeners, that is not the case, the payload is actually:

`{
    "job": "Illuminate\\Events\\CallQueuedHandler@call",
    "data": {
        "class": "App\\Listeners\\Frontend\\Auditfile\\AuditfileUploadedListener",
        "method": "handle",
        "data": "a:1:{i:0;O:47:\"App\\Events\\Frontend\\Auditfile\\AuditfileUploaded\":1:{s:58:\"\u0000App\\Events\\Frontend\\Auditfile\\AuditfileUploaded\u0000auditfile\";O:45:\"Illuminate\\Contracts\\Database\\ModelIdentifier\":2:{s:5:\"class\";s:30:\"App\\Models\\Auditfile\\Auditfile\";s:2:\"id\";i:4;}}}"
    }
}`

Compare that to the payload of a normal Job:
`{"job":"Illuminate\\Queue\\CallQueuedHandler@call","data":{"commandName":"App\\Jobs\\ImportAuditfileJob","command":"O:27:\"App\\Jobs\\ImportAuditfileJob\":5:{s:38:\"\u0000App\\Jobs\\ImportAuditfileJob\u0000auditfile\";O:45:\"Illuminate\\Contracts\\Database\\ModelIdentifier\":2:{s:5:\"class\";s:30:\"App\\Models\\Auditfile\\Auditfile\";s:2:\"id\";i:9;}s:10:\"connection\";N;s:5:\"queue\";N;s:5:\"delay\";N;s:6:\"\u0000*\u0000job\";N;}"}}`

I hope this is enough information

vladimir-yuldashev:
Yes. In 5.2 already.

nanophp:
> Yes. In 5.2 already.

@vyuldashev I'm using Laravel 6.6 but I have the same problem.

vyuldashev:
@nanophp Package version?

nanophp:
> @nanophp Package version?

the package version is 9.1.2  and laravel version is 6.6.0

[...]

PR #67, opened 2016-07-12
Move attempts data into job metadata, fixes for some Job types

dustindoiron:
Jobs like the standard mail queue don't have a `$body['data']['command']` array key, causing an `ErrorException` upon attempting to work them.

This alters the strategy for placing attempt logic into the AMQP message metadata, and safely handles switching between standard Jobs with objects present at the command key, and jobs that don't have this key present.


dustindoiron:
(This also resolves #66)

vladimir-yuldashev:
Thanks!

Issue #68, opened 2016-07-19
Add support for Correlational ID


CaporalDead:
Hi @vladimir-yuldashev,

I'm trying to add to laravel a way to support "correlational id" (as you can see here https://github.com/Devitek/laravel-correlational-id) and I'd like to add this feature to message queue with your package. You can find an example of this here https://www.sitepoint.com/php-rabbitmq-advanced-examples/. Do you have an idea to implement this feature with your package ? I will try to find a good way on my side but if you're interested by that let me know :).

vladimir-yuldashev:
Why do we need that?

CaporalDead:
The concept of the correlational ID is a convenient way to trace messages or log across multiple applications, For example, if you work on a micro service architecture with multiple APIs or using a queue and if you're using an ELK stack maybe, you can easily follow your request accross your APIs or daemons to find what's going on.

vladimir-yuldashev:
Done thx to @dustindoiron.

CaporalDead:
Thx @dustindoiron @vladimir-yuldashev !!! Didn't have the time to look @ it,

PR #73, opened 2016-08-09
Backports correlation/job ID support from master into v5.2

dustindoiron:
Backports the correlation id support from master into v5.2.

 * Tested on laravel/framework:v5.2.39

### DIFF
--- a/src/VladimirYuldashev/LaravelQueueRabbitMQ/Queue/Jobs/RabbitMQJob.php
+++ b/src/VladimirYuldashev/LaravelQueueRabbitMQ/Queue/Jobs/RabbitMQJob.php
@@ -160,10 +160,26 @@ class RabbitMQJob extends Job implements JobContract
     /**
      * Get the job identifier.
      *
-     * @return string
+     * @return string|false
      */
     public function getJobId()
     {
-        return $this->message->get('correlation_id');
+        if ($this->message->has('correlation_id') === true) {
+            return $this->message->get('correlation_id');
+        }
+
+        return false;
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


src/VladimirYuldashev/LaravelQueueRabbitMQ/Queue/Jobs/RabbitMQJob.php
```diff
@@ -160,10 +160,26 @@ private function setAttempts($count)
     /**
      * Get the job identifier.
      *
-     * @return string
+     * @return string|false
```
vladimir-yuldashev: It's not acceptable.

vladimir-yuldashev:
It's the question about `correlation_id` already solved?

https://github.com/vladimir-yuldashev/laravel-queue-rabbitmq/blob/5.2/src/VladimirYuldashev/LaravelQueueRabbitMQ/Queue/Jobs/RabbitMQJob.php#L167

src/VladimirYuldashev/LaravelQueueRabbitMQ/Queue/Jobs/RabbitMQJob.php
dustindoiron: Fixed

dustindoiron:
Two things with the current implementation:
 * `correlation_id` is set by the application, see https://www.rabbitmq.com/amqp-0-9-1-reference.html
 * since `correlation_id` is not set, it throws an `OutOfBoundsException` in GenericContent (from the PHP amqplib library).

This should resolve both technical issues, as well as #68 by providing support for it.

PR #74, opened 2016-08-10
[5.1] Allow custom properties when publishing a job

SteffenBach:
RabbitMQ supports additional properties on the message beyond the delivery mode and content type. These properties can be used for sending metadata along the published payload, which can be very useful if the message consumer is written in something other than Laravel.

### DIFF
--- a/src/VladimirYuldashev/LaravelQueueRabbitMQ/Queue/RabbitMQQueue.php
+++ b/src/VladimirYuldashev/LaravelQueueRabbitMQ/Queue/RabbitMQQueue.php
@@ -71,11 +71,17 @@ class RabbitMQQueue extends Queue implements QueueContract
 			$queue = $this->declareDelayedQueue($queue, $options['delay']);
 		}
 
-		// push job to a queue
-		$message = new AMQPMessage($payload, [
+		$properties = [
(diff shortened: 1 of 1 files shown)


SteffenBach:
If I'm understanding you correctly, you want to rename the array key `properties` to something else. 
I'm not against that, but I chose the term _properties_ very deliberately. The reason being, it matches the terminology used by the RabbitMQ Management UI when viewing a message.

I don't see the conflict you are talking about. Would you mind elaborating?

vladimir-yuldashev:
@SteffenBach I mean get properties from incoming ``$options['_rabbitmq_properties']`` and convert to ``$properties['properties']``

SteffenBach:
@vladimir-yuldashev But that introduces an extra level of nesting, which prevents these properties from being really useful. They should be allowed to operate at the root level, so that it's possible for the caller to override what is specified in `$properties`. I.e., if they want to override the content type, or specify one of the other RabbitMQ message properties.

vladimir-yuldashev:
@SteffenBach sorry about forgetting this PR. Is this still required?

beerkeeper:
+1 for this feature.
I'm looking for a way to give messages priority over the other, and with this solution I can do that.

PR #81, opened 2016-08-19
Convert DateTime to seconds

mortenhauberg:
When using a DateTime or Carbon object in the later method.


mortenhauberg:
Apparently the `.travis.yml` file is missing and StyleCI is pretty grumpy.
What would you like me to do here :)?

PR #84, opened 2016-08-19
Convert DateTime to seconds

mortenhauberg:
When using a DateTime or Carbon object in the later method.


mortenhauberg:
I'm a bit confused.
Does StyleCI wants me to remove a `return null;`?

I can remove it, no problem, but how did it end up there in the first place :)?

PR #85, opened 2016-08-19
Convert DateTime to seconds

mortenhauberg:
When using a DateTime or Carbon object in the later method:

Sorry for the noise. I messed up my PR ;)


mortenhauberg:
Would you like me to remove the `return null;`, that's making StyleCI grumpy?

vladimir-yuldashev:
@mortenhauberg StyleCI will make his job himself. ;)

PR #90, opened 2016-09-09
optimize bind function

reatang:
Solves the queue to send too slow


vladimir-yuldashev:
We already have such implementation in 5.2 and 5.3. Just need to move code to 5.1 too.

PR #94, opened 2016-10-17
Added better error handling to rabbitmq connectors

samlev:
If the RabbitMQ connector was unavailable, and the queue was running in daemon mode, the error log would flood within seconds. This reports the connection error responsibly, and sleeps to stop the log file from being instantly flooded.

This is against the 5.1 branch, as I'm using that branch (because of Laravel LTS being 5.1), but I am happy to also send a similar update to the 5.2 and/or 5.3 branches.

### DIFF
--- a/src/VladimirYuldashev/LaravelQueueRabbitMQ/Queue/RabbitMQQueue.php
+++ b/src/VladimirYuldashev/LaravelQueueRabbitMQ/Queue/RabbitMQQueue.php
@@ -3,8 +3,11 @@
 namespace VladimirYuldashev\LaravelQueueRabbitMQ\Queue;
 
 use DateTime;
+use ErrorException;
+use Exception;
 use Illuminate\Contracts\Queue\Queue as QueueContract;
 use Illuminate\Queue\Queue;
+use Log;
 use PhpAmqpLib\Channel\AMQPChannel;
 use PhpAmqpLib\Connection\AMQPConnection;
 use PhpAmqpLib\Message\AMQPMessage;
@@ -23,6 +26,7 @@ class RabbitMQQueue extends Queue implements QueueContract
 	protected $defaultQueue;
 	protected $configQueue;
 	protected $configExchange;
+	protected $sleepOnError;
 
 	/**
 	 * @param AMQPConnection $amqpConnection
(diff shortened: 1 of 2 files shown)


vladimir-yuldashev:
You need to handle errors in your application.

samlev:
The point is that you can't handle these errors. This is rabbitmq not responding, so the error log gets flooded in seconds.

vladimir-yuldashev:
I have another idea. Custom exception and it will be thrown upon error. Developers will decide how to react on that. Sleep is not good.

samlev:
A catch-able exception is already being thrown, but there's no-where in the application where it can be caught. This happens in `artisan queue:work --daemon`. The process is *meant* to run something like this something like this:

> **Laravel core Code**
> Begin work
> Load adaptor

>> **This package**
>> Connect to Rabbit MQ
>> Find the next job to run

>>> **Application code**
>>> Run job

>> Clean up

> Clean up
> Go back to step 1

In reality, what is happening is this:

> **Laravel core Code**
> Begin work
> Load adaptor

>> **This package**
>> Connect to Rabbit MQ
>> _Failed to connect! Throw exception!_

> Catch Exception and write to log file
> Go back to step 1

The problem is that application code doesn't get called at all, and can't catch the exception. This package can't halt execution of the 'work' process, because even if it did, a new process would (and is) started up immediately.

Because there is no delay, the exception gets written to the log file repeatedly until the disk runs out of space (a matter of seconds to minutes to hours, depending on the size of the machine this is running on, and how many queue processors are being run). Seeing as an outage of rabbitmq is unlikely to be fixed in the milliseconds it takes for the next process to retry, we just flood the error log with failure messages, and have no way to handle this from within the application code itself.

That's why I put a delay in - the rabbitmq server is more likely to be available again after 3 seconds than 0.03, and even if it's not, that's even more reason not to hammer it in the hope that it's there. This doesn't resolve the *core* problem, but it does help to alleviate the symptom that makes the server crash.

vladimir-yuldashev:
Okay, I got you. I will merge your pull request and make small changes. Thanks!

vladimir-yuldashev:
@samlev Okay. I applied changes to 5.1 release. Can you update package in your project and confirm that changes work?

samlev:
Updated code to use this new version, and tested breaking the rabbitmq connection - appears to work as intended.

vladimir-yuldashev:
@samlev Nice, thanks.

PR #97, opened 2016-12-06
[5.3] Fix wrong implements of default attempt count

sweetvvck:


vladimir-yuldashev:
Nice. Thanks for contribution.

PR #106, opened 2017-02-20
Add closing connection feature (in previous version old connection wa…

StanislavPivovarov:
…s still alive  despite for worker was die or reload and create new for himself )


StanislavPivovarov:
I change some logic, and move it to this new branch. I thing it's will be better to hold this functionality in provider. For test I test it and it's ok. But in travis 5.6 it fail how  I see in log it's not related to my changes

StanislavPivovarov:
Would you apply this request?

vladimir-yuldashev:
@StanislavPivovarov Merged. Thanks.

PR #108, opened 2017-02-22
[bugfix] laravel 5.4 now parses jobs uses JobName static method parse

pdscopes:
`laravel/framework v5.4.12` no longer uses `$this->parseJob()`. Updated to use `JobName::parse()`


vladimir-yuldashev:
Updated 5.4 tag, thanks!

PR #118, opened 2017-05-16
BugFix - Added missing connectionName.

sazo:
When calling $this->fail() inside a RabbitMQ job it fails because the connectionName is NULL.

I have looked at how core laravel jobs does it and this is how. :)

`
Next exception 'Illuminate\Database\QueryException' with message 'SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'connection' cannot be null (SQL: insert into failed_jobs (connection, queue, payload, exception, failed_at) .....
`


sazo:
@vladimir-yuldashev will you move the 5.4 tag to include this?

vladimir-yuldashev:
@sazo Yes, a bit later as soon I will get to my computer. Did you test that driver still works? 😂

sazo:
@vladimir-yuldashev Np - yeah i tested it on your development environment ;)

sazo:
love the driver btw! :)

vladimir-yuldashev:
@sazo Released.

PR #125, opened 2017-06-29
Fix for lost connections with RabbitMQ

werewolf81:
At the moment, if a queue worker is running while a connection error happens to RabbitMQ, the worker becomes corrupted and cannot process data correctly anymore.

The job being run by the worker is never marked as failed, it also does not work anymore even if the connection to RabbitMQ is restored.

This fix allows to optionally throw an exception if this scenario happens and the worker will be killed by Laravel.
Then if used along with Supervisor, a new worker is spawned and new connection will be established allowing the system to work again.

### DIFF
--- a/src/VladimirYuldashev/LaravelQueueRabbitMQ/Queue/RabbitMQQueue.php
+++ b/src/VladimirYuldashev/LaravelQueueRabbitMQ/Queue/RabbitMQQueue.php
@@ -332,7 +332,7 @@ class RabbitMQQueue extends Queue implements QueueContract
 
         // If it's set to false, throw an error rather than waiting
         if ($this->sleepOnError === false) {
-            throw new \Exception('Error writing data to the connection with RabbitMQ');
+            throw new \RuntimeException('Error writing data to the connection with RabbitMQ');
         }
 
         // Sleep so that we don't flood the log file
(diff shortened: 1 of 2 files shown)


src/VladimirYuldashev/LaravelQueueRabbitMQ/Queue/RabbitMQQueue.php
```diff
@@ -324,10 +324,17 @@ public function getCorrelationId()
...
      * @param Exception $e
+     * @throws Exception
      */
     protected function reportConnectionError($action, Exception $e)
     {
         Log::error('AMQP error while attempting '.$action.': '.$e->getMessage());
+
+        // If it's set to false, throw an error rather than waiting
+        if ($this->sleepOnError === false) {
+            throw new \Exception('Error writing data to the connection with RabbitMQ');
```
vladimir-yuldashev: Probably changing Exception to RuntimeException will make better sense?

werewolf81: Sure, that's fine. I've applied the change.

vladimir-yuldashev:
Thanks, merged!

PR #135, opened 2017-08-08
Queue params

RatkoR:
This PR adds config for additional queue arguments.

This arguments can be any of: `x-message-ttl`, `x-max-length`, ... See [docs](https://www.rabbitmq.com/maxlength.html) for these options. They are also listed in rabbitmq management web under the "Create new queue" command.

Since this options are key/value pairs I choose json notation in `.env` file (ENV() doesnt support arrays, just strings).

Please comment if there's anything you want me to change. I've changed tests and they all pass ok.

### DIFF
--- a/config/rabbitmq.php
+++ b/config/rabbitmq.php
@@ -39,6 +39,7 @@ return [
         'durable'     => env('RABBITMQ_QUEUE_DURABLE', true),
         'exclusive'   => env('RABBITMQ_QUEUE_EXCLUSIVE', false),
         'auto_delete' => env('RABBITMQ_QUEUE_AUTODELETE', false),
+        'arguments'   => env('RABBITMQ_QUEUE_ARGUMENTS', false),
     ],
     'exchange_params' => [
         'name' => env('RABBITMQ_EXCHANGE_NAME', null),
--- a/config/rabbitmq.php
+++ b/config/rabbitmq.php
@@ -39,7 +39,7 @@ return [
         'durable'     => env('RABBITMQ_QUEUE_DURABLE', true),
         'exclusive'   => env('RABBITMQ_QUEUE_EXCLUSIVE', false),
         'auto_delete' => env('RABBITMQ_QUEUE_AUTODELETE', false),
-        'arguments'   => json_decode(env('RABBITMQ_QUEUE_ARGUMENTS', '{}'), 1),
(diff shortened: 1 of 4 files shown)


config/rabbitmq.php
```diff
@@ -39,6 +39,7 @@
         'durable'     => env('RABBITMQ_QUEUE_DURABLE', true),
         'exclusive'   => env('RABBITMQ_QUEUE_EXCLUSIVE', false),
         'auto_delete' => env('RABBITMQ_QUEUE_AUTODELETE', false),
+        'arguments'   => json_decode(env('RABBITMQ_QUEUE_ARGUMENTS', '{}'), 1),
```
vyuldashev: Did you try using `json_decode` with `config:cache`. Does it still work?

RatkoR: Yes, it does work. At least here for our project. Does it not work for you?

RatkoR: I'm looking at the failed travis now...

vyuldashev: It just looks strange to `json_decode` string in configuration. I suggest to move it to `RabbitMQQueue` constructor. What do you think?

tests/RabbitMQQueueTest.php
```diff
@@ -80,7 +82,9 @@ public function test_push()
             $this->config['queue_params']['passive'],
             $this->config['queue_params']['durable'],
             $this->config['queue_params']['exclusive'],
-            $this->config['queue_params']['auto_delete']
+            $this->config['queue_params']['auto_delete'],
+            false,
+            Mockery::any()
```
RatkoR: Can you find a better way to test here?

If I add `new AMQPTable($this->config['queue_params']['arguments'])` it doesn't pass. Probably because objects are different. I could make it pass any other way but with `any()`...

config/rabbitmq.php
RatkoR: Updated. I added a new variable `$configQueueArguments`. I don't know how you feel about adding it, since I could just: 

`$this->configQueue['arguments'] = json_decode($this->configQueue['arguments'], 1)`

..  it didn't seem ok because it overwrites the original value. So I opted to create a new variable. If you want it differently just say, I'm ok with either (or something else that you prefer).

Did you see my comment in tests?

PR #138, opened 2017-08-31
Added SSL support for the package.

werewolf81:
It took me a while to figure out how to make Travis work with my changes. 
There's a bug in RabbitMQ which requires to have 2 identical config files (with an additional .config appendix) if you want to use custom configs so in the build there's a line which makes a copy of the main file.

Also, for some reason RabbitMQ doesn't want to use custom configs if started by:
sudo service rabbitmq-server start
so it must be started manually as a background process.

Sudo is required to stop the original RabbitMQ service so I had to remove that line which disables sudo from the Travis configuration file.

### DIFF
--- a/config/rabbitmq.php
+++ b/config/rabbitmq.php
@@ -60,10 +60,10 @@ return [
      */
     'ssl_params' => [
         'ssl_on'        => env('RABBITMQ_SSL', false),
-        'cafile'        => env('SSL_CAFILE', null),
-        'local_cert'    => env('SSL_LOCALCERT', null),
-        'verify_peer'   => env('SSL_VERIFY_PEER', true),
-        'passphrase'    => env('SSL_PASSPHRASE', null),
+        'cafile'        => env('RABBITMQ_SSL_CAFILE', null),
+        'local_cert'    => env('RABBITMQ_SSL_LOCALCERT', null),
+        'verify_peer'   => env('RABBITMQ_SSL_VERIFY_PEER', true),
+        'passphrase'    => env('RABBITMQ_SSL_PASSPHRASE', null),
     ]
 
 ];
--- a/config/rabbitmq.php
+++ b/config/rabbitmq.php
@@ -55,4 +55,15 @@ return [
      */
     'sleep_on_error' => env('RABBITMQ_ERROR_SLEEP', 5),
 
+    /*
+     * Optional SSL params if an SSL connection is used
+     */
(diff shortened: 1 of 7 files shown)


werewolf81:
Does anyone know why this build fails whereas it passes in my fork?
https://travis-ci.org/werewolf81/laravel-queue-rabbitmq/builds/270404669

For some reason, RabbitMQ doesn't read or like my SSL configuration in the original package.

werewolf81:
Ok, so the problem was somehow caused by using the 'precise' environment instead of the 'trusty' one.

vyuldashev:
What about `xenial`? `precise` is EOL already.

werewolf81:
So it looks like Xenial isn't quite supported yet by Travis, however, it accepts this distribution in the config file.
`
Build system information
Build language: php
Build group: stable
**Build dist: xenial** (my setting from .travis.yml)
Build id: 270736044
Job id: 270736045
Runtime kernel version: 4.4.0-83-generic
travis-build version: 826b19c20
Build image provisioning date and time
Fri Jul  7 18:51:38 UTC 2017
Operating System Details
Distributor ID:	Ubuntu
Description:	**Ubuntu 14.04.5 LTS** (the actual OS used)
Release:	14.04
Codename:	**trusty**
`

config/rabbitmq.php
```diff
@@ -55,4 +55,15 @@
      */
     'sleep_on_error' => env('RABBITMQ_ERROR_SLEEP', 5),
 
+    /*
+     * Optional SSL params if an SSL connection is used
+     */
+    'ssl_params' => [
+        'ssl_on'        => env('RABBITMQ_SSL', false),
+        'cafile'        => env('SSL_CAFILE', null),
```
vyuldashev: Always add `RABBITMQ` prefix to all variable names

werewolf81: The reason why I didn't do it there was the the same setting can be used for SSL connections to the database and other services. 
So once you start using SSL, the path to the CA file doesn't have to be related just to RabbitMQ.

vyuldashev: We are not eligible to make changes that can affect other parts of framework. So, this option should be strictly to this library.

werewolf81: That makes sense. I'll change it and update the pull request.

vyuldashev:
Great job. I will test it and push to 5.5 release later.

PR #142, opened 2017-09-07
Update RabbitMQJob.php

lboulay:
Add a return statement when you release a job.
This return contains the new correlationId.


vyuldashev:
According to the base Job, release method should return void. https://github.com/laravel/framework/blob/5.3/src/Illuminate/Queue/Jobs/Job.php#L111

PR #152, opened 2017-10-21
Add a new option connection_method_socket for gaining better performance

axot:
Compare to `AMQPSocketConnection` `AMQPStreamConnection` get much slower response on first connection.

https://github.com/php-amqplib/php-amqplib/issues/516
https://github.com/php-amqplib/php-amqplib/issues/449

### DIFF
--- a/config/rabbitmq.php
+++ b/config/rabbitmq.php
@@ -56,6 +56,12 @@ return [
     'sleep_on_error' => env('RABBITMQ_ERROR_SLEEP', 5),
 
     /*
+     * Optional php-amqplib param, AMQPSocketConnection could gain better performance.
+     * If set to false, AMQPStreamConnection will be used (default).
+     */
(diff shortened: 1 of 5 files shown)


axot:
Is it possible backport this to all versions?

vyuldashev:
We need to give developer an option to use `AMQPStreamConnection` or `AMQPSockerConnection`

axot:
I agree your idea, a new option `connection_method_socket` was added ;D

vyuldashev:
I think configuration value should be called something like `connector` and it will be an array with class name and options. 

```php
'connector' => [
    'class' => RabbitMQConnector::class,
    'options' => [],
],
```

Class can be one of `RabbitMQConnectorSocket::class`, `RabbitMQConnectorSSL::class` or `RabbitMQConnector::class` by default.

axot:
Should we move `ssl_params` into new `connector -> options`, what do you think?

vyuldashev:
See #154

PR #157, opened 2017-11-08
Refactoring Configuration

vyuldashev:


vyuldashev:
@makasim Is dsn really required? Can't we just use host, port, vhost options, etc.?

makasim:
It is not required at all. You can use either options or dsn string. You can even combine both (dsn options have higher priority).

makasim:
LGTM

PR #154, opened 2017-10-26
Amqp interop based version.

makasim:
ref https://github.com/vyuldashev/laravel-queue-rabbitmq/issues/153

This is far from the final version, the aim of this PR is to show differences in approaches.

PR #158, opened 2017-11-10
[doc] add instruction on how to change the transport.

makasim:
addition to https://github.com/vyuldashev/laravel-queue-rabbitmq/pull/154


makasim:
@vyuldashev could you please have a look at this too?

PR #164, opened 2017-12-14
add connection name to Job

net53:


vyuldashev:
Why do we need this?

net53:
For example:
\Illuminate\Queue\InteractsWithQueue::fail

PR #167, opened 2018-01-02
Fix chained jobs attempts

net53:
There are Job1 and Job2.
Added to the queue using Job1::withChain([new Job2 ()])->dispatch();

If Job1 $tries = 2 and Job2 $tries = 1. Then if the second attempt is successful for Job1, Job2 will not be executed because $ job->attempts () for Job2 will immediately return 2.

And if $tries are the same, then the number of attempts for the next will be less by the number of failures in the chain than should be.


net53:
And another case. 
If Job2 was dispatched from Job1 on second attempt, then Job2 will fail because of attempts_count=2.

Please check and merge this important fix.

Issue #165, opened 2017-12-16
receiveNoWait vs receive


leonrenkema:
I'm trying this module and when I run the queue worker with the following command 
`artisan queue:work --queue=medium --sleep=0` the load is going up and I think because Rabbitmq is overloaded with polling requests. 

The following line is causing this problem because this wont wait until a new message can come available `if ($message = $consumer->receiveNoWait()) {` in RabbitMQ.php:115
When I change this line to `if ($message = $consumer->receive(10)) {` the polling overload is not happening because the call blocks for 10 seconds.

I'm not sure if this can be fixed or made optional in this module to choose for blocking or non blocking mode. 

For performance reasons I think blocking would be better than polling. And for Rabbitmq this makes more sense.

makasim:
`$message = $consumer->receive(10)` - 10 here is 10 milliseconds not 10 seconds.

makasim:
I think the behavior should be configurable. So a developer can decide on what method to use.

leonrenkema:
It is indeed milliseconds, did not see that. The exact timeout should also be configurable I guess.

PR #172, opened 2018-01-16
Use basic consume

deweller:
By using basic_consume, a queue worker can respond very quickly to a queued message.

This does not poll the rabbitmq server.  Instead it blocks with an open connection to the rabbitmq server.

This is a solution for #165.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
     "illuminate/database": "5.5.*",
     "illuminate/support": "5.5.*",
     "illuminate/queue": "5.5.*",
-    "enqueue/amqp-lib": "0.8.5",
+    "enqueue/amqp-lib": "^0.8.15",
     "queue-interop/amqp-interop": "^0.7"
   },
   "require-dev": {
--- a/config/rabbitmq.php
+++ b/config/rabbitmq.php
@@ -93,4 +93,15 @@ return [
         'passphrase' => env('RABBITMQ_SSL_PASSPHRASE', null),
     ],
 
+    'receive' => [
+        /**
+         * Use basic_consume by default which will block and not poll
+         */
+        'method' => env('RABBITMQ_RECEIVE_METHOD', 'basic_consume'),
+
+        /**
+         * The timeout (in milliseconds) to block and wait for a message
+         */
+        'timeout' => env('RABBITMQ_RECEIVE_TIMEOUT', 5000),
+    ],
 ];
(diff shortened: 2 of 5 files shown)


deweller:
The only failing test is this:

```
1) VladimirYuldashev\LaravelQueueRabbitMQ\Tests\Functional\SslConnectionTest::testConnectorEstablishSecureConnectionWithRabbitMQBroker
ErrorException: stream_socket_client(): unable to connect to ssl://127.0.0.1:5671 (Connection refused)
```

That is not related to this pull request.

lanphan:
It's around 1 year now.
@deweller May I support you to create new PR for this?
@vyuldashev Is there anything required to update?

composer.json
```diff
@@ -13,7 +13,7 @@
     "illuminate/database": "5.5.*",
     "illuminate/support": "5.5.*",
     "illuminate/queue": "5.5.*",
-    "enqueue/amqp-lib": "0.8.5",
```
vyuldashev: Let’s stick with 0.8.* for now. 0.9 has breaking changes.

config/rabbitmq.php
```diff
@@ -93,4 +93,15 @@
         'passphrase' => env('RABBITMQ_SSL_PASSPHRASE', null),
     ],
 
+    'receive' => [
```
vyuldashev: Probably better place for this configuration is `options.connection` or `options.worker`?

vyuldashev:
@lanphan I just submitted comments and PR should be updated from master.

config/rabbitmq.php
lanphan: I prefer options.connection, as it has timeout setting.

composer.json
lanphan: OK, I saw your comment somewhere about this.

src/Queue/Connectors/RabbitMQConnector.php
```diff
@@ -57,6 +57,7 @@ public function connect(array $config): Queue
             'ssl_cert' => $config['ssl_params']['local_cert'],
             'ssl_key' => $config['ssl_params']['local_key'],
             'ssl_passphrase' => $config['ssl_params']['passphrase'],
+            'receive_method' => isset($config['receive']) ? $config['receive']['method'] : 'basic_get',
```
makasim: fyi: the option is removed in 0.9

src/Queue/RabbitMQQueue.php
```diff
@@ -123,11 +129,18 @@ public function pop($queueName = null)
...
-            list($queue) = $this->declareEverything($queueName);
+            list($queue) = $this->declareEverythingOnce($queueName);
 
-            $consumer = $this->context->createConsumer($queue);
+            // create the consumer once and cache it
+            $consumer = $this->createConsumerOnce($queue);
 
-            if ($message = $consumer->receiveNoWait()) {
+            if (isset($this->receiveConfig['method']) and $this->receiveConfig['method'] == 'basic_consume') {
+                $message = $consumer->receive($this->receiveConfig['timeout']);
```
makasim: fyi: the method using basic get under the hood in 0.9. To use basic consume a subscription consumer should be used

lanphan: thanks, I'll read more docs about this

PR #184, opened 2018-04-21
support the publish rabbitmq config file

kinyou:
support exec php artisan vendor:publish command copy the rabbitmq.php to app/config/rabbitmq.php


vladkucherov:
👍

vyuldashev:
I don't like the idea to create separate configuration file (`rabbitmq.php`), its better to put values into your `queue.php` file.

PR #185, opened 2018-05-10
Adds support for custom headers and properties on pushRaw

vladkucherov:


webernir:
Thanks. 
super helpful!

PR #188, opened 2018-05-25
[6.0][7.0] Added config publishes and rabbitmq lazy connector.

ed-fruty:
This PR adds: 
 - Config publishes feature when running in console mode.
 - RabbitMQ connector initialization only when queue is resolved, no force queue manager initialization more. If it was already initialized - connector will add right now.


vyuldashev:
There is no need to publish and/or merge configuration file.

PR #192, opened 2018-06-19
Allow user to configure job class to be created in queue

davidstanley01:
This PR allows an end user to specify a custom job class to be created when processing messages off the queue.

For my use-case, we have a fanout queue that distributes messages created by other systems (some PHP systems, some Java systems, etc). By specifying a custom class, I can easily process those messages without having them change their format to fit the Laravel/Lumen way of doing things.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ Change the factory class in `config/queue.php`:
     ],
 ```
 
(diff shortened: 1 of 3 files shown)


davidstanley01:
Any interest in this at all?

vyuldashev:
Hello, I think you can just extend current driver and make your changes.

PR #199, opened 2018-08-15
added configuration option for delay strategy class

mostafasaeed88:
added configuration option for delay strategy class and wrote it's tests to make sure it's working properly

### DIFF
--- a/config/rabbitmq.php
+++ b/config/rabbitmq.php
@@ -18,7 +18,7 @@ return [
      *  - \EnqueueAmqpBunny\AmqpConnectionFactory if you install enqueue/amqp-bunny
      */
     'factory_class' => Enqueue\AmqpLib\AmqpConnectionFactory::class,
-
+    'delay_strategy_class' => Enqueue\AmqpTools\RabbitMqDlxDelayStrategy::class,
     'host' => env('RABBITMQ_HOST', '127.0.0.1'),
     'port' => env('RABBITMQ_PORT', 5672),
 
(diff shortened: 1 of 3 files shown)


config/rabbitmq.php
```diff
@@ -18,7 +18,7 @@
      *  - \EnqueueAmqpBunny\AmqpConnectionFactory if you install enqueue/amqp-bunny
      */
     'factory_class' => Enqueue\AmqpLib\AmqpConnectionFactory::class,
-
+    'delay_strategy_class' => Enqueue\AmqpTools\RabbitMqDlxDelayStrategy::class,
```
vyuldashev: Can we just name it `delay_strategy`?

src/Queue/Connectors/RabbitMQConnector.php
```diff
@@ -42,7 +43,15 @@ public function connect(array $config): Queue
         if (false === class_exists($factoryClass) || false === (new \ReflectionClass($factoryClass))->implementsInterface(InteropAmqpConnectionFactory::class)) {
             throw new \LogicException(sprintf('The factory_class option has to be valid class that implements "%s"', InteropAmqpConnectionFactory::class));
         }
-
+        
+        $delayStrategyClass = RabbitMqDlxDelayStrategy::class;
+        if (array_key_exists('delay_strategy_class', $config)) {
```
vyuldashev: Since we are requiring php 7.1, you can use 
`$delayStrategyC = $config['delay_strategy'] ?? RabbitMqDlxDelayStrategy::class;

src/Queue/Connectors/RabbitMQConnector.php
```diff
@@ -42,7 +43,15 @@ public function connect(array $config): Queue
...
+             $delayStrategyClass = $config['delay_strategy_class'];
+            if (false === class_exists($delayStrategyClass) || false === (new \ReflectionClass($delayStrategyClass))->implementsInterface(DelayStrategy::class)) {
+                throw new \LogicException(sprintf('The delay_strategy_class option has to be valid class that implements "%s"', DelayStrategy::class));
```
vyuldashev: no need to use reflection. just use instanceof

PR #216, opened 2019-01-12
support basic_consume and basic_get connection

lanphan:
Thanks to PR #172 from @deweller, I create new PR based on latest master branch, please help to review again.

1. support basic_consume and basic_get connection
2. use cache for consumer and declarations
3. fix unit test


vyuldashev:
Does current version uses `basic_get`?

vyuldashev:
@makasim mentioned:

> fyi: the method using basic get under the hood in 0.9. To use basic consume a subscription consumer should be used

I guess it's better to upgrade to enqueue 0.9, firstly.

PR #218, opened 2019-02-01
Clean settings from rabbitmq in config/queue.php

robertoarruda:
Merge the `config/rabbitmq.php` in lib with `config/queue.php`


vyuldashev:
I already closed similar issues. I don't think it is good to merge config into `config/queue.php`. You have to do it manually.

Issue #232, opened 2019-03-17
Ability to add a backoff strategy to the delay strategy.


adm-bome:
When a message has failed and should be retried with a delay, this library is already handling this via the `Enqueue\AmqpTools\RabbitMqDlxDelayStrategy`.

This strategy is usefull in many situations, But if you want a more advanced scenario like described in this [article](https://m.alphasights.com/exponential-backoff-with-rabbitmq-78386b9bec81), you can't use this library because the chosen strategy just is not capable to do it.

As stated in the [article](https://m.alphasights.com/exponential-backoff-with-rabbitmq-78386b9bec81) rabbitMQ does not play well with mixed ttl`s in the same DLX queue.

It would be nice to add the ability to manage a delay strategy from config and also add the abbitity to choose a backoff strategy.

adm-bome:
@vyuldashev what do you think of this enhancement? I think this makes the library more flexible. Also this feature is a compliment to how RabbitMQ works with DLX messages.

adm-bome:
This feature was intended for the V9 or lower versions...
Sind V10 ( more inline with Queue API from Laravel :+1: ) 

Delaying messages with multiple delaingTimes (Backoff) should not be build in into this library. 
When someone wants to alter the DelayStrategy this must be handled with hooks on events, thrown by the worker.

Also see this [issue](https://github.com/laravel/framework/issues/30509)

PR #233, opened 2019-03-17
Add possibility to change delay strategy

adm-bome:
#232 Possible solution for this issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ Once you completed the configuration you can use Laravel Queue API. If you used 
 
 ## Laravel Horizon Usage
 
-Starting with 7.4, this package supports [Laravel Horizon](http://horizon.laravel.com) out of the box. Firstly, install Horizon and then set `RABBITMQ_WORKER` to `horizon`.
+Starting with 8.0, this package supports [Laravel Horizon](http://horizon.laravel.com) out of the box. Firstly, install Horizon and then set `RABBITMQ_WORKER` to `horizon`.
 
 ## Lumen Usage
 


src/Queue/Connectors/RabbitMQConnector.php
```diff
@@ -69,11 +69,11 @@ public function connect(array $config): Queue
 
         if ($factory instanceof DelayStrategyAware) {
             /** @var DelayStrategy $delayStrategy */
-            $delayStrategyClass = Arr::get($config, 'delay.strategy', RabbitMqDlxDelayStrategy::class);
+            $delayStrategyClass = Arr::get($config, 'delay.strategy') ?? RabbitMqDlxDelayStrategy::class;
             $delayStrategy = new $delayStrategyClass();
 
             if ($delayStrategy instanceof BackoffStrategyAware) {
-                $backoffStrategyClass = Arr::get($config, 'delay.backoff.strategy', ConstantBackoffStrategy::class);
+                $backoffStrategyClass = Arr::get($config, 'delay.backoff.strategy') ?? ConstantBackoffStrategy::class;
```
vyuldashev: You can pass ConstantBackoffStrategy::class as a third argument to Arr::get

adm-bome: Thanx for the feed back. 

I did that, but when the config item exists and is null, null will be returned. So that's wy i changed it to ??

vyuldashev: Third option in Arr::get is exactly for the default value.

adm-bome: I know. But `Arr::get` returns the value if the key exists, even if it is set to `null`. When initiating a ne Class of `null` you get an error.

Because the config contains defaults and the key's exist, `null` is always returned when not using this feature.

adm-bome: I updated the branch.

When...

1. The keys in the config are omitted. 
2. The values for the keys in the config are `null`.
3. The values for the keys in the config aren't existing classes with a correct interface.

...the default for that specific Awareness kicks in.

What ever happens in the config, the library will work without errors and is working like it was before (backwards compatible).

adm-bome:
#232 #233 Ready for review or merge.

> This feature is compatible with 7.* and 8.* versions of this library

vyuldashev:
This is huge. I guess we need tests for this.

adm-bome:
Hopefully not 🤔.
I know it looks huge. But the changes are minimal. 

The Connector has code to test if config values are as expected. If not then it uses defaults. 

By using interfaces and testing on them, it should be ok without tests.

PR #238, opened 2019-03-28
Feature: Alter the delay-strategy from config.

adm-bome:
- manage the delay-strategy from config.
 - add backoff-strategy on delayed messages.
 - add prioritization of delayed messages.

### DIFF
--- a/config/rabbitmq.php
+++ b/config/rabbitmq.php
@@ -24,6 +24,46 @@ return [
      */
     'factory_class' => Enqueue\AmqpLib\AmqpConnectionFactory::class,
 
+    /*
+     * ## Manage the delay strategy from the config.
+     *
+     * The delay strategy can be set to:
+     *  - \VladimirYuldashev\LaravelQueueRabbitMQ\Queue\Tools\DlxDelayStrategy::class
+     *
+     * ### Backoff Strategy
+     *
+     * The `DlxDelayStrategy` is BackoffAware and by default a ConstantBackoffStrategy is assigned.
+     * This ensures the same behavior as if the `RabbitMqDlxDelayStrategy` was assigned.
+     *
+     * You can assign different backoffStrategies with extra options, for example:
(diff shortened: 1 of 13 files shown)


adm-bome:
#232

config/rabbitmq.php
```diff
@@ -24,6 +24,46 @@
      */
     'factory_class' => Enqueue\AmqpLib\AmqpConnectionFactory::class,
 
+    /*
```
vyuldashev: Lets put this whole block as a last key of config

vyuldashev:
Its a huge one. The feature is interesting. I need some time to go through it. Also, tests would be useful.

vyuldashev:
How will this strategy work with this: https://github.com/vyuldashev/laravel-queue-rabbitmq/blob/master/src/Queue/RabbitMQQueue.php#L117?

config/rabbitmq.php
adm-bome: Ok. I will move this to the end

adm-bome:
As intended.
The delay is in seconds in laravel. So when there was an delay set in an command/worker, milliseconds are set into the Producer. This feature does not change this behavior but follows it.


This Feature only adds the ability to choose between DelayStrategy's, as stated in the config. The way the code currently works, was not changed. The feature hooks into this.

Giving developers the ability to:
- do nothing, and use the thirth party RabbitMqDlxDelayStrategy (default).
- choose there own DelayStrategy (when they write one).
- choose the already available `DlxDelayStrategy` in this repo (added).

In case of the `DlxDelayStrategy` developers can:
- add an **optional** BackoffStrategy.
- configure an **optional** Prioritization for messages when they where delayed multiple times.

> When RabbitMq is going to handle multiple messages with an delay, and the delay is increasing by the times the messages were delayed (backoff), then the delayed messages can not stay into one DLX queue, because rabbitMQ does not handle this correctly (by design). (also see: the link in #232)

PR #243, opened 2019-05-13
queue: add message routingKey property

hex7c0:
set proper `routingKey` if is different from the queue name

### DIFF
--- a/src/Queue/RabbitMQQueue.php
+++ b/src/Queue/RabbitMQQueue.php
@@ -72,14 +72,33 @@ class RabbitMQQueue extends Queue implements QueueContract
              */
             [$queue, $topic] = $this->declareEverything($queueName);
 
+            /** @var AmqpMessage $message */
             $message = $this->context->createMessage($payload);
-            $message->setRoutingKey($queue->getQueueName());
+
             $message->setCorrelationId($this->getCorrelationId());
             $message->setContentType('application/json');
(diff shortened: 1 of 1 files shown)


src/Queue/RabbitMQQueue.php
```diff
@@ -72,16 +72,35 @@ public function pushRaw($payload, $queueName = null, array $options = [])
...
                 $message->setExpiration($options['expiration']);
             }
 
+            if (isset($options['delivery_tag'])) {
+                $message->setDeliveryTag($options['delivery_tag']);
+            }
+
+            if (isset($options['consumer_tag'])) {
+                $message->setConsumerTag($options['consumer_tag']);
+            }
```
makasim: What is the reason for setting consumer_tag and deliver_tag here? They make sense for the consumer but not producer.

makasim: Also, they should not be set manually, the broker sets them

hex7c0: well, since the message have a lot of configuration/property, I'd like to put them all to avoid other PRs, like my previous one
what do you think?

makasim: A developer should not set delivery_tag or consumer_tag never. They are set by the broker only.

hex7c0: fine for me

vyuldashev: Why these methods even exists then?

PR #247, opened 2019-05-24
queue: accept custom RabbitMQQueue class

hex7c0:
Add the ability to change Queue class

queue.php configuration file can be
`'worker' => \App\Providers\Foo::class,`

and Foo class should be something like

```php
/**
 * Class Foo.
 */
class Foo extends RabbitMQQueue
{
}
```

### DIFF
--- a/src/Queue/Connectors/RabbitMQConnector.php
+++ b/src/Queue/Connectors/RabbitMQConnector.php
@@ -86,6 +86,10 @@ class RabbitMQConnector implements ConnectorInterface
             return new HorizonRabbitMQQueue($context, $config);
         }
 
+        if (class_exists($worker)) {
+            return new $worker($context, $config);
+        }
+
         throw new InvalidArgumentException('Invalid worker.');
     }
 }


hex7c0:
hey @vyuldashev, any news?

vyuldashev:
Released this feature v8.3.0. Class should be instance of `RabbitMQQueue`

hex7c0:
man I'm sorry, how is possible that your code work if I put this in the configuration file?

'worker' => \App\Providers\Foo::class,

vyuldashev:
@hex7c0 You call should be instance of RabbitMQQueue class from this package.

hex7c0:
yes but in the configuration queue.php file I just set my class

*if* I have to instance the class, I have to pass `AmqpContext $context, array $config` repeating your code and I think that is unnecessary duplication

what do you think?

anolek:
Agree with @hex7c0 

Instead of using "instanceof", we could use "is_subclass_of()" and this will work fine (for string classname and instance of RabbitMQQueue)

hex7c0:
gentle ping

PR #251, opened 2019-06-10
Implemented basic_consume

Zuken:
Implemented ability to consume by basic_consume instead of polling.
References to #216


vyuldashev:
I have plans to add a separate `rabbitmq:consume` artisan command. Currently closing this PR, but will refer to it when adding a feature.

PR #271, opened 2019-10-03
Updated configuration for providers

chillichicken:


vyuldashev:
If you disable auto-discovery you should be aware of registering service provider manually.

PR #276, opened 2019-10-21
Fixed getJobId for v8

eyaylagul:


vyuldashev:
This package does not support v8 as long as Laravel 5.8 support finished at August 26th, 2019.

PR #278, opened 2019-10-21
Fix #277

### LINKED ISSUE #277: Unable to use default exchange name
The default RabbitMQ exchange name (an empty string) can't be used.

When an empty string is supplied as the exchange name, the library will attempt to create a new exchange with the same name as the queue.

garbetjie:
Attempt to fix issue https://github.com/vyuldashev/laravel-queue-rabbitmq/issues/277.

Only defaults to the queue name if a non-string is supplied for the exchange name (ie: supplying `null` will use the queue name as the exchange name).


vyuldashev:
Will be fixed in v10.

PR #280, opened 2019-11-15
Add reconnect feature.

orlissenberg:
We had some problems with the queue worker(s) because they would got into "Broken pipe or closed connection" error flooding mode when the RabbitMQ cluster was restarted for maintenance.

We are currently fixing that by letting the (managed) workers crash by setting the RABBITMQ_ERROR_SLEEP=false Because they are managed they will start again when the cluster is back up.

A reconnect feature would be nicer, hence this PR.


vyuldashev:
Sorry I do not accept PRs to old versions of library.

orlissenberg:
Would you like me to add one for the master (lastest) version or do you have another/better solution for this issue?

vyuldashev:
Well, I think it is totally okay that worker will stop on error.

orlissenberg:
The problem is that it won't stop and malfunctions if the default setting of RABBITMQ_ERROR_SLEEP=5 is used, it will not reconnect properly, to reproduce:
- start rabbitmq
- start the queue worker
- stop and start rabbitmq
- log gets flooded with "AMQP error while attempting pop: Broken pipe or closed connection"
- queue worker does not stop

also happens on the ^9.1

src/Queue/RabbitMQQueue.php
```diff
@@ -131,10 +138,26 @@ public function pop($queueName = null)
                 return new RabbitMQJob($this->container, $this, $consumer, $message);
             }
         } catch (\Exception $exception) {
+            if ($exception instanceof AMQPConnectionClosedException ||
+                $exception instanceof AMQPChannelClosedException
```
vyuldashev: This can be in a separate catch block:

```
catch(AMQPConnectionClosedException | AMQPChannelClosedException $exception) {
    // ...
}
```

vyuldashev:
Well, I think we can remove `reportConnectionError` method, which was introduced in #94 and I think this method is not relevant in latest Laravel versions, we will also remove try-catch blocks so worker will stop and try to reconnect. What do you think?

src/Queue/RabbitMQQueue.php
orlissenberg: Fixed and I added a PR for the master branch also.

orlissenberg:
> Well, I think we can remove `reportConnectionError` method, which was introduced in #94 and I think this method is not relevant in latest Laravel versions, we will also remove try-catch blocks so worker will stop and try to reconnect. What do you think?

It looks like Beanstalkd tries to reconnect a single time (without delay) but would indeed stop after and would depend on an external process to restart the worker. For our setup that could be up to a 30 minute delay before the worker is up and running again (Puppet run interval).

Would be nice if a (delayed) (single/multiple) reconnect option would still be configurable (but not default), not sure if everyone using the library wants to setup supervisord, Puppet, systemd, etc.
Removing the ```reportConnectionError ``` sounds like a good idea. Would make it an announced feature change.

vyuldashev:
> It looks like Beanstalkd tries to reconnect a single time (without delay) but would indeed stop after and would depend on an external process to restart the worker. For our setup that could be up to a 30 minute delay before the worker is up and running again (Puppet run interval).

That's a correct behaviour. We should accept it too.

vyuldashev:
> not sure if everyone using the library wants to setup supervisord, Puppet, systemd, etc.

How do you run your workers then?

orlissenberg:
Managed by systemd & Puppet

This is a part from the systemd unit configuration:

```
[Service]
ExecStart=/usr/bin/env php /example_project_path/releases/current_root/artisan queue:work rabbitmq --queue=example_queue --tries=3 --daemon
Type=simple
User=example_user
Group=example_group
Restart=on-failure
RestartSec=5
StartLimitBurst=3
StartLimitInterval=120s
```

vyuldashev:
I will not introduce reconnect feature in v10. Exceptions will be thrown and process manager will need to restart workers. This is correct behavior of workers.

orlissenberg:
> I will not introduce reconnect feature in v10. Exceptions will be thrown and process manager will need to restart workers. This is correct behavior of workers.

I'll close this PR, you already merged/fixed the master branch, thanks!! 👍

Issue #235, opened 2019-03-23
Forever retry when job timeout


ngyuki:
hi maintainer.

I have a following job class.

```php
<?php

namespace App\Jobs;

use Illuminate\Bus\Queueable;
use Illuminate\Queue\SerializesModels;
use Illuminate\Queue\InteractsWithQueue;
use Illuminate\Contracts\Queue\ShouldQueue;
use Illuminate\Foundation\Bus\Dispatchable;

class SampleJob implements ShouldQueue
{
    use Dispatchable, InteractsWithQueue, Queueable, SerializesModels;

    public function handle()
    {
        sleep(10);
    }
}
```

When job timeout, it ignores `--tries` option and retries forever.

```console
$ while :; do php artisan queue:work --delay=1 --sleep=1 --tries=3 --timeout=3 rabbitmq; done
[2019-03-23 00:14:19][5c957a5b1b0b04.32221024] Processing: App\Jobs\SampleJob
Killed
[2019-03-23 00:14:24][5c957a5b1b0b04.32221024] Processing: App\Jobs\SampleJob
Killed
[2019-03-23 00:14:29][5c957a5b1b0b04.32221024] Processing: App\Jobs\SampleJob
Killed
[2019-03-23 00:14:35][5c957a5b1b0b04.32221024] Processing: App\Jobs\SampleJob
Killed
[2019-03-23 00:14:40][5c957a5b1b0b04.32221024] Processing: App\Jobs\SampleJob
Killed
[2019-03-23 00:14:45][5c957a5b1b0b04.32221024] Processing: App\Jobs\SampleJob
Killed
[2019-03-23 00:14:50][5c957a5b1b0b04.32221024] Processing: App\Jobs\SampleJob
Killed
[2019-03-23 00:14:55][5c957a5b1b0b04.32221024] Processing: App\Jobs\SampleJob
Killed
[2019-03-23 00:15:01][5c957a5b1b0b04.32221024] Processing: App\Jobs\SampleJob
Killed
[2019-03-23 00:15:06][5c957a5b1b0b04.32221024] Processing: App\Jobs\SampleJob
Killed
:
:
:
```

In case of redis driver,  job will fails after the number of `--tries` option.

```console
$ while :; do php artisan queue:work --delay=1 --sleep=1 --tries=3 --timeout=3 redis; done
[...]

PR #281, opened 2019-11-15
Fixing attempts incrementing:

mrvtoney:
In the RabbitMQQueue.php:
- $message was changed to a class property ($this->message)

In the RabbitMQJob.php:
- class const ATTEMPT_COUNT_HEADERS_KEY was changed from “attempts_count” to “attempts”

### DIFF
--- a/src/Queue/Jobs/RabbitMQJob.php
+++ b/src/Queue/Jobs/RabbitMQJob.php
@@ -21,7 +21,7 @@ class RabbitMQJob extends Job implements JobContract
     /**
      * Same as RabbitMQQueue, used for attempt counts.
      */
-    public const ATTEMPT_COUNT_HEADERS_KEY = 'attempts';
+    public const ATTEMPT_COUNT_HEADERS_KEY = 'attempts_count';
 
     protected $connection;
     protected $consumer;
--- a/src/Queue/Jobs/RabbitMQJob.php
+++ b/src/Queue/Jobs/RabbitMQJob.php
@@ -21,7 +21,7 @@ class RabbitMQJob extends Job implements JobContract
     /**
      * Same as RabbitMQQueue, used for attempt counts.
      */
-    public const ATTEMPT_COUNT_HEADERS_KEY = 'attempts_count';
+    public const ATTEMPT_COUNT_HEADERS_KEY = 'attempts';
 
     protected $connection;
     protected $consumer;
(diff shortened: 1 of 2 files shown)


mrvtoney:
@vyuldashev I can't figure out what errored out for the styleci/pr check. Can you point it out for me?

vyuldashev:
> $message was changed to a class property ($this->message)
I don't think it's a good idea. Why did you add this?

mrvtoney:
The message was a class property in the RabbitMQJob as well.

However, simply changing the value of the const ATTEMPT_COUNT_HEADERS_KEY did not fix the problem as a whole.

The message's state needed to persist through the process in order to increment. Every time the message was used it was recreated but when it was recreated. It did not take into consideration or apply the logic for the attempts. Meaning that each time a new message was created and the attempts were set to one again instead of incrementing(if at all).

vyuldashev:
I just [pushed](https://github.com/vyuldashev/laravel-queue-rabbitmq/commit/d7dafddbbd303c75ac63e6b016cc0dd369f27f96) tests for job attempts. And seems like it works like expected: https://travis-ci.org/vyuldashev/laravel-queue-rabbitmq/jobs/613081677

Also, we are using this library in many application in production which somehow have logic with attempts count and everything works.

mrvtoney:
That is great to hear.

Yeah. While going through the laravel worker code. I saw that database, redis and sync all utilized attempts.

I modified it in order to maintain consistency with laravel's other queue drivers.

I can change it back and test on my side to see if the change from attempt_counts to attempts.

mrvtoney:
@vyuldashev I tested and changed 'attempts' back to 'attempts_count'

vyuldashev:
So why do I need to accept this PR if attempts are incrementing?

mrvtoney:
Attempts do not increment without the changes to RabbitMQQueue.php.

The attempt_count isnt incrementing. It fixes this issue:

https://github.com/vyuldashev/laravel-queue-rabbitmq/issues/235

mrvtoney:
@vyuldashev do you have any other suggestions for my PR?

vyuldashev:
I will review and test it and probably this goes to v10

vyuldashev:
I tried your code, seems like it does not fix #235

mrvtoney:
@vyuldashev Can you send me the steps you used to test so I can test again on my side, please?

vyuldashev:
@mrvtoney I used the same code from #235 
I think I found a solution but it will break BC. So, I need to work on it and it will go to v10

PR #293, opened 2019-12-18
Exchange type should be an option, not argument

MorrisonHotel:
In the Console/ExchangeDeclareCommand.php exchange type declared as an option, but used as argument.

```
tm@mo:/var/www$ php artisan rabbitmq:exchange-declare test --type=topic

The "type" argument does not exist.
```


vyuldashev:
Thank you!

PR #296, opened 2019-12-30
Allow to disable non-blocking mode

MorrisonHotel:
With non-blocking mode we have CPU usage ~100%.
So, lets declare an option for `rabbitmq:consume` command to toggle this behavior.

Actualy, I'd rather to set up `non_blocking=false` while channel is waiting, by default. But, maybe it's a BC breaking change.

`php artisan rabbitmq:consume` - current behavior (with non-blocking mode)
`php artisan rabbitmq:consume --blocking` - with blocking mode

### DIFF
--- a/src/Console/ConsumeCommand.php
+++ b/src/Console/ConsumeCommand.php
@@ -37,7 +37,7 @@ class ConsumeCommand extends WorkCommand
         $consumer->setConsumerTag($this->consumerTag());
         $consumer->setPrefetchSize((int) $this->option('prefetch-size'));
         $consumer->setPrefetchCount((int) $this->option('prefetch-count'));
-        $consumer->setNonBlocking(!$this->option('blocking'));
+        $consumer->setNonBlocking(! $this->option('blocking'));
 
         parent::handle();
     }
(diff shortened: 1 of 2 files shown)


MorrisonHotel:
@vyuldashev ping

vyuldashev:
Blocking mode is not a solution because timeouts will not work. I will provide fix based on [this](https://github.com/SynergiTech/laravel-queue-rabbitmq/pull/1) approach.

vyuldashev:
Fixed in v10.1.3.

SlayerBirden:
> Blocking mode is not a solution because timeouts will not work. I will provide fix based on [this](https://github.com/SynergiTech/laravel-queue-rabbitmq/pull/1) approach.

@vyuldashev  I think you're wrong on this one. Blocking mode is exactly the mode where timeouts are actually used. See https://github.com/php-amqplib/php-amqplib/blob/1252cfbab6342dd36f5c5f1af70b6085bfcf60f8/PhpAmqpLib/Channel/AbstractChannel.php#L345

There's no point in timeouts for non-blocking mode, because it's not actually waiting.

vyuldashev:
@SlayerBirden by timeout I mean `timeout` parameter in `rabbitmq:consume` command and timeout in job.

SlayerBirden:
@vyuldashev in that case I'm not sure how it's connected to blocking mode. 

As I understand it blocking mode affects how `channel->wait` works (limits how long we're waiting for a message), while timeout in rabbitmq:consume affects Laravel's `runJob` (limits how long job works).

PR #298, opened 2020-01-12
Missing Container class fix

mohov:


adm-bome:
To old and probably not actual anymore

Issue #301, opened 2020-01-17
[10.x] Dead Letter exchange flags cause errors when message is rejected


SlayerBirden:
- Laravel Version: 6.x
- Package Version: 10.1.3
- RabbitMQ Version: 3.6.16

**Describe the bug**
Probably connected to #299. `pushRaw` now declares queue with 2 additional options: x-dead-letter-exchange and x-dead-letter-routing-key. However it specifies the queue itself as the value for those.
Based on the doc (https://www.rabbitmq.com/dlx.html) the message would be routed using exchange/key to another queue(s) if it's "nack'd", failed because of timeout etc. Not sure what was the idea behind assigning queue itself as a destination. It results in errors when messages are rejected due to Exceptions: something like 

*operation basic.reject caused a channel exception precondition_failed: unknown delivery tag 1*.

What happens - is after **reject** message is redirected to Dead Letter exchange - which is... the same queue. So it's consumed again resulting in that error.

I don't think Dead Letter exchange functionality should be part of "default" setup. Otherwise it needs to be implemented properly.

**Steps To Reproduce**
Just dispatch new message to RabbitMq.
**Expected behavior**
Dead letter exchange is not used.

adm-bome:
@SlayerBirden 

Please verify your issue is resolved with the latest commits
`composer require vyuldashev/laravel-queue-rabbitmq:dev-master`

SlayerBirden:
hi @adm-bome sorry for the late response. We've switched to the older release and I don't have time to test it at the moment. I expect we'll try out 10.x release at some point, and I'll update this issue.

vyuldashev:
Should be fixed in v10.2.0

Issue #299, opened 2020-01-13
[10.x] Unable to work with queue


codercms:
- Laravel Version: 6.x
- Package Version: 10.1.3
- RabbitMQ Version: 3.7.16

**Describe the bug**
1. Unable to run `php artisan rabbitmq:consume rabbitmq --queue=...` or ` php artisan queue:work --queue=...` commands without declaring the queue.
1. When using `php artisan rabbitmq:queue-declare` command, there is no way to push any job using pushRaw method of this package, because `queue-declare` command declaring the queue in different way than pushRaw.

**Steps To Reproduce**
Try to run:
```
php artisan rabbitmq:consume rabbitmq --queue=some
```
It will result in error (`PhpAmqpLib\Exception\AMQPProtocolChannelException  : NOT_FOUND - no queue 'some' in vhost '/'`).

Let's declare the queue:
```
php artisan rabbitmq:queue-declare some
```

And run the queue worker again:
```
php artisan rabbitmq:consume rabbitmq --queue=some
```
Now it's running.

Job class:
```php
class Some implements ShouldQueue
{
    public function fire(RabbitMQJob $job, array $payload): void
    {
        $job->release(5);  
    }
}
```

Trying to dispatch a job through RabbitMQ Management:
```json
{"job": "App\\Jobs\\Some", "data": {}}
```

The job execution will result in error (because of "release" method):
```
PRECONDITION_FAILED - inequivalent arg 'x-dead-letter-exchange' for queue 'some' in vhost '/': received the value 'some' of type 'longstr' but current is none
```
And after that worker falls to an infinite loop with this message: `Channel connection is closed.`

Trying to dispatch the same job from CLI:
```php
Queue::pushRaw(json_encode(['job' => 'App\\Jobs\\Some', 'data' => []]), 'some');
```
It will result in error:
```
PhpAmqpLib/Exception/AMQPProtocolChannelException with message 'PRECONDITION_FAILED - inequivalent arg 'x-dead-letter-exchange' for queue 'some' in vhost '/': received the value 'some' of type 'longstr' but current is none'
```

**Expected behavior**
Everything is works like in 9.x

**Additional context**

adm-bome:
@codercms 

Please verify your issue is resolved with the latest commits
`composer require vyuldashev/laravel-queue-rabbitmq:dev-master`

codercms:
@adm-bome I can confirm it works now.
[...]

PR #303, opened 2020-01-22
Fix queue not declaring while setting worker and consumer

josembarboteo:
#299 
#301

### DIFF
--- a/src/Queue/RabbitMQQueue.php
+++ b/src/Queue/RabbitMQQueue.php
@@ -201,6 +201,10 @@ class RabbitMQQueue extends Queue implements QueueContract
         try {
             $queue = $this->getQueue($queue);
 
+            $this->declareExchange($queue);
+            $this->declareQueue($queue, true, false);
+            $this->bindQueue($queue, $queue, $queue);
+
             /** @var AMQPMessage|null $message */
             if ($message = $this->channel->basic_get($queue)) {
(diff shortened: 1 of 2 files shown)


src/Queue/RabbitMQQueue.php
```diff
@@ -204,6 +201,10 @@ public function pop($queue = null)
         try {
             $queue = $this->getQueue($queue);
 
+            $this->declareExchange($queue);
+            $this->declareQueue($queue, true, false);
+            $this->bindQueue($queue, $queue, $queue);
+
```
adm-bome: Same as my previous comment.

A Consumer `$this->connection->pop('queue')` should not create exchage, queue or bindings.

I know there is an issue:
When a queue does not exits, the active channel is closed. The consumer is looping and can't get messages.
Resulting in an worker/consumer loop. (Death worker)

To solve this issue see: [Link](https://github.com/adm-bome/laravel-queue-rabbitmq/blob/8a21c99bb7ba54da3a48f7196e3f27279a7eae79/src/Queue/RabbitMQQueue.php#L256)

adm-bome: maybe you can up vote my pull request #304 to solve the issues

adm-bome:
I think this PR can be closed, after the release of v10.2.0

PR #304, opened 2020-02-02
Fix laravel job attempts, when removing DLX

adm-bome:
- Fixed: losing messages when no delay was set
- Fixed: the updating of attempts when pushing the job back to the queue.
- Removed: unnecessary/wrong options on queues
- Added: auto cleanup of unused delay queue's
- Added prioritization for failed jobs

#305


src/Queue/Jobs/RabbitMQJob.php
```diff
@@ -75,15 +75,14 @@ public function attempts(): int
         $headers = Arr::get($this->message->get_properties(), 'application_headers');
 
         if (! $headers) {
-            return 0;
+            return 1;
```
vyuldashev: By default, all Laravel queue drivers return 0 for first attempt.

src/Queue/Jobs/RabbitMQJob.php
```diff
@@ -75,15 +75,14 @@ public function attempts(): int
...
         }
 
         $data = $headers->getNativeData();
 
         $laravelAttempts = (int) Arr::get($data, 'laravel.attempts', 0);
-        $xDeathCount = (int) Arr::get($headers->getNativeData(), 'x-death.0.count', 0);
 
-        return $laravelAttempts + $xDeathCount;
+        return $laravelAttempts + 1;
```
vyuldashev: x-death count is used when job has been rejected, nacked or requeued. It should be taken into consideration

src/Queue/Jobs/RabbitMQJob.php
```diff
@@ -108,15 +107,9 @@ public function delete(): void
      */
     public function release($delay = 0): void
     {
-        parent::release($delay);
+        parent::release();
 
-        if ($delay > 0) {
```
vyuldashev: This was an intended behaviour in order attempts to work as expected.

src/Queue/Jobs/RabbitMQJob.php
adm-bome: This is true when pushing a job for the first time to the queue. 
When a cosumer/worker gets the message for the first time attempts() must return the current try. 

Sea also the RedisJob of the framework:

[Link](https://github.com/jimrubenstein/laravel-framework/blob/f7ba54c132319631163ac0c98101da0a3b16c3ca/src/Illuminate/Queue/Jobs/RedisJob.php#L108)

src/Queue/Jobs/RabbitMQJob.php
adm-bome: My solution does not care about nacked messages. Ofcourse you have to tell rabbitMQ what to do with the message, when the job is released. But because the DLX was removed on the main Queue, rabbitMQ does nothing with the message and the message is gone after nacking or acknowledging.

My solution always pushes a 'new' message with different attributes/properties to the queue.

See also my comment in issue: #305

src/Queue/Jobs/RabbitMQJob.php
adm-bome: Because my solution always pushes a 'new' message to the queue by calling the laterRaw(). When the delay is zero or in the past, the 'new' message is pushed with pushRaw() 

[See](https://github.com/adm-bome/laravel-queue-rabbitmq/blob/849d608be1ddfe81f25dbe2f49f238d38cda6054/src/Queue/RabbitMQQueue.php#L146)

In your solution a failed message was rejected/nacked and so always in front of the queue, withholding other messages to get proccesed. In my solution the message gets a higher priority, over other messages, to get processed first.

adm-bome: Maybe there is an solution for loosing the messages when no DLX is set and rejecting messages, and keeping the xDeatCount property into account.

I have to test this situation.

CASE:
define queue without DLX.

When rejecting also requeue the message, by calling `$this->reject($this, true)`.

Hopefully xDeathCount is updated when nacking the message. 

The only problem i think persistst: is that when there are 1000 Jobs in the queue, the failed Job will proccess only when the other 999 are processed first. Leaving us with a situation that we can't tell when in time the message is processed again.

src/Queue/Jobs/RabbitMQJob.php
adm-bome: Maybe there is an solution for not loosing the messages, when no DLX is set and rejecting messages and keeping the xDeathCount property into account.

I have to test this situation.

CASE:
define queue without DLX.

When rejecting also requeue the message, by calling `$this->reject($this, true)`.

Hopefully xDeathCount is updated when nacking the message. 

The only problem i think persistst: is that when there are 1000 Jobs in the queue, the failed Job will proccess only when the other 999 are processed first. Leaving us with a situation that we can't tell when in time the message is processed again.

adm-bome:
Conclusion:

- I will revert some things. ;)
- DLX on main queue, must be removed.
- Message attribute laravel.attempts must be updated when working with delay.
- message property priority must be included when working with delay.

The attempts() method is correct.

src/Queue/Jobs/RabbitMQJob.php
adm-bome: result:  
xDeathCount is only set when DLX is used. 
DLX set to its own queue results in errors. DLX was not designed to route messages to it self (as far as i know).

When removing DLX on the main queue (in your code) we cant use `reject()`, or message will be lost.

src/Queue/Jobs/RabbitMQJob.php
adm-bome: Because I removed DLX on the main queue, and always publish a new message... this intention was removed.
`laravel.attempts` in the message arguments is always updated accordingly and xDeathCount was rejected.

I think this conversation can be closed.

vyuldashev:
First of all, I want to thank you for the huge work. I need some time to review everything.

README.md
```diff
@@ -73,6 +73,94 @@ Add connection to `config/queue.php`:
...
+
+    'rabbitmq' => [
+        // ...
+
+        'options' => [
+            'queue' => [
+                // ...
+
+                'prioritize_delayed_messages' =>  false,
+                'queue_max_priority' => 10,
```
vyuldashev: Is it possible to every job has it's own priority in a queue?

adm-bome: Yes this is possible.

When we delay messages in RabbitMQ by sending them to a queue with DLX, the message (when expires) is put back at the **end** of the queue.

Case: 
When the queue holds more than a 1000 messages (Jobs), the delayed job is processed again after the sum of (delayTime + (processingTime * 1000)).

To make sure the delayed job is processed as soon as it comes out of the delayQueue, a higher priority is set and rabbitMQ makes sure the delayed message is transferred to the **front** of the queue.

When working with one delayTime (one DelayQueue), 2 priority levels are sufficient. When implementing a delayStrategy (multiple delayQueues), you probably want more priority levels.
This to make sure the oldest Job is processed first or as soon as possible.

adm-bome:
> First of all, I want to thank you for the huge work. I need some time to review everything.

vyuldashev:
Let's make all changes incremental. Fix bugs then add new functionality in separate PRs. Huge PRs make hard to make decision to merge them.

README.md
vyuldashev: May we we can add it in every job. Like through properties or even method.

adm-bome:
> Let's make all changes incremental. Fix bugs then add new functionality in separate PRs. Huge PRs make hard to make decision to merge them.

This is a rework. The issues people have can't be solved with the current code. 
The rework does exactly the same thing as it did before. except the strategy has changed. All tests should prove that. 

I added some features because the previous versions supported these. This because people/developers who were using those features cant upgrade at the moment.

vyuldashev:
I mean, I am willing to merge fixes sooner than features.

README.md
adm-bome: Prioritizing Jobs from within a Job can be done by dispatching the job to an other queue. (hig priority queues. (Laravel mentions this in the documentation.)

When a job fails, u can dispatch the Job to a queue which will be processed first by the workers. there is a method `failed()` u can use for this.

Maybe i miss understood the question: But i don't think such a feature should be part of this library.

adm-bome:
Oke, take your time ;)

We cant fix anything with the current codebase (no disrespect).
The strategy is wrong. I realy did my research ;)
I tried do do make as minimal changes as possible.

vyuldashev:
I don't want to add job's priorities and setting custom exchanges at the moment. May be in a separate PRs.

adm-bome:
No wories... You don't have to merge anything.
I have a fork ;).

It would be nice if we all could use 'one' library. But for now i will use my own.

vyuldashev:
Closing for now. Will provide fixes based on this PR.

SchizoDuckie:
@adm-bome are you planning to make your lib a hard fork? I'm trying to orient myself on the right package to use and your research definitely needs to be part of it.

adm-bome:
> @adm-bome are you planning to make your lib a hard fork? I'm trying to orient myself on the right package to use and your research definitely needs to be part of it.

@SchizoDuckie :
No need. This package is the right package, if u want to work with Laravel / Queue API.

PR #317, opened 2020-02-25
code clean-up

adm-bome:
- refactored some duplicated code
- altered some logic to get the values from message.
- acknowledge message in the tests when done with job.
- wrote the tests for the queue arguments.
- added exchange/queue delete commands with tests


vyuldashev:
@adm-bome Can you DM me on twitter or Telegram? Same username.

PR #321, opened 2020-03-05
Fallback connection

Dezinger:
Add feature "fallback" connection. If current rabbitmq connection is failed we redirect job to another connection.


adm-bome:
- No support for this version ;)
- Checks fail

PR #322, opened 2020-03-12
When a job fails it tries to reject it twice.

phalouvas:
When a job fails in Consumer.php the override of function markJobAsFailedIfWillExceedMaxAttempts causes job to try to reject twice resulting failure on the consumer, due to false tag message.

This occurs only if the job running throw exception for some reason. Above fix remove that function. After that it works as expected.

### DIFF
--- a/src/Consumer.php
+++ b/src/Consumer.php
@@ -154,23 +154,6 @@ class Consumer extends Worker
     }
 
     /**
-     * Mark the given job as failed if it has exceeded the maximum allowed attempts.
-     *
-     * @param string $connectionName
-     * @param Job|RabbitMQJob $job
-     * @param int $maxTries
-     * @param Exception $e
-     */
-    protected function markJobAsFailedIfWillExceedMaxAttempts($connectionName, $job, $maxTries, $e): void
-    {
-        parent::markJobAsFailedIfWillExceedMaxAttempts($connectionName, $job, $maxTries, $e);
-
-        if (! $job->isDeletedOrReleased()) {
-            $job->getRabbitMQ()->reject($job);
-        }
-    }
-
-    /**
      * Stop listening and bail out of the script.
      *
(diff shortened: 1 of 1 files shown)


vyuldashev:
Try v10.2.0

phalouvas:
Hello,

 

I transferred the core Joomla data but there is problem with the users. You have many users in the source site with no username. That is not allowed in the new Joomla version.

 

Can you please check how many users like this you have in the source site?

 

Kind regards, Panayiotis Halouvas

adm-bome:
@vyuldashev @phalouvas I think this can be closed

phalouvas:
Did not manage to test so far, but you can close it since it passed all your tests. If I find something in the future, will let you know.

adm-bome:
As of v10.2.0 the library strategy is different. I think this issue is resolved.

May you still find yourself having this issue, dont hesitate to mention this in this issue.

phalouvas:
Unfortunately problem still exists with new version 10.2.

if you start the consumer with `php artisan rabbitmq:consume --queue=default --tries=3`
and a job throws exception, it does not rerun. It gives this: https://1drv.ms/u/s!AljvtJPILTC7gZ_LM-EJmK5XartGhQg?e=C0K38m

adm-bome:
I have one question, when removing this method, what happens with the message in RabbitMQ?
Because an error occurs, the message must not be lost within the process.
It must either be released or deleted. 

> As the message can't be handled twice: Where did it go or what happened to the message?

@vyuldashev 
Was there a special case why these extra lines were added into this method?

phalouvas:
@adm-bome if I remember correctly the job is removed. Not 100% sure though. Need to test again to see. In any case the whole thing needs revision.

adm-bome:
I will look into the consumer part, somewere in the next 3 days.

PR #332, opened 2020-04-21
Updated dependencies for Laravel 7

Norgul:


valehasadli:
Anyone tested?

Norgul:
Well I did...I wouldn't submit a pull request otherwise :)

adm-bome:
There is already support for 7+

Norgul:
It wasn't there when I was making modifications

adm-bome:
In the composer file there is the minimum of versions. Updating to just laravel 7 dependicies will break support for  laravel 6.

When everyting works like expected. Maybe you can close this PR.

valehasadli:
> In the composer file there is the minimum of versions. Updating to just laravel 7 dependicies will break support for  laravel 6.
> 
> 
> 
> When everyting works like expected. Maybe you can close this PR.
> 
> 

What is your offer in this case? Laravel day by day improving versions, and package not compatible with newer versions.

valehasadli:
> Well I did...I wouldn't submit a pull request otherwise :)

I hope merge will accept soon)

Norgul:
> In the composer file there is the minimum of versions. Updating to just laravel 7 dependicies will break support for laravel 6.

That is kind of a whole point of versioning stuff. You will make a new version and say that it is not backwards compatible. If someone needs L6 support, he will use older version of the repo in the same way he is using older version of Laravel

adm-bome:
@Norgul 

> You will make a new version and say that it is not backwards compatible.

Why?
If you make breaking changes in the code of the package, I agree with you. A package can be compatible with more than one version of the main project (in this case Laravel) and when Laravel won't drastically change the queue Api concept there is no need to make a different version. 

If your suggesting to follow the convention of Laravel and just make a new version for every time a new major version of Laravel is released, Then I say: This is pointless and not the point of versioning and even time consuming because you have to maintain all different versions.

Maybe i'm not getting what you are trying to achieve here. I can install this package in laravel 6 and 7 without errors and everything works fine. Just making a version 11.0.0 for Laravel 7 is crazy.

adm-bome:
@valehasadli 

> Laravel day by day improving versions, and package not compatible with newer versions.

Of course Laravel will improve and so will this library. 
I can install this package in laravel 6 and 7 without errors and everything works fine.
Why do you need these specific laravel 7 changes?

Norgul:
It didn't work for me on a brand new L7 project, otherwise I wouldn't have a reason to update dependencies. 

Major version is called major for a reason. It includes major changes which may or may not be backwards compatible. If it were, then it would exist no reason to call it major, and you could simply bump the minor version number

adm-bome:
Just did the following

- Installed/created a brand new laravel v7.13.0
- Installed this library v10.2.1
- started rabbitMQ
- started SQL server
- configured laravel to use the database and rabbitMQ as queue connection
- created a TestJob
- dispatched the TestJob
- consumed a the TestJob


**All without any errors.**

Norgul:
I suppose others looking at this PR had the same issues. Maybe some other package is clashing with it. I just made the updates which ultimately work for me, so I will let the project owner decide on what to do. My fork works for me, and if it does for others then great.

adm-bome:
There is indeed nothing wrong with the pull-request. 

Except it will break support for laravel/lumen 6.

I have no problems installing and using this package in laravel/lumen 7.
So I don't see why this PR has to be merged.

Until somebody can point out **why** this package wont install, or **why** this package doesn't work or can't be used in laravel/lumen 7, this PR will not be merged and no major version will be made.

vyuldashev:
@Norgul Hello. Can you elaborate on what exactly is not working for you with Laravel 7.0 exactly?

Norgul:
I honestly have no idea. When I was trying to install it a month ago it broke because of the dependency mismatch. It isn't a stock L7 project for sure (I tried it once again and it indeed works). Today I have no idea which project was it so I can't reproduce it to know where it broke. I will close the request. If anyone tries and it breaks, let them handle it or submit a new issue/PR.

Issue #338, opened 2020-05-28
Listen for non-job formatted messages (general messages)


etudor:
This may seem like a stupid question, but how can I listen for general messages?   
I have messages published on the rabbitmq queue from another application and I would like to handle them via a consumer/job handler/worker or anything. 

I managed to publish messages to the rabbitmq with the library so the connection is properly made. 
I am using laravel 7.13 (latest) and the latest version of this package as well.

This may not be necessarily a question for this package or Laravel framework.

Could I create an old type rabbitmq consumer? I don't necessarily need to create a job handler.
I would assume what I want to do is quite easy but haven't found the right answer yet.

I see that the same question is on StackOverflow but the answer is not very clear and I don't want to install another rabbitmq laravel package. https://stackoverflow.com/questions/32505501/consume-raw-json-queue-messages-in-laravel

Many thanks

adm-bome:
What do you mean by a **general message**? 

Most of laravel consumers worker work the way laravel creates a payload. 
A typical payload would be:

```json
{
  "uuid": "6616c749-89a1-4d74-acfb-4ec26eb49794",
  "displayName": "App\\Jobs\\TestJob",
  "job": "Illuminate\\Queue\\CallQueuedHandler@call",
  "maxTries": null,
  "maxExceptions": null,
  "delay": null,
  "timeout": null,
  "timeoutAt": null,
  "data": {
    "commandName": "App\\Jobs\\TestJob",
    "command": "O:16:\"App\\Jobs\\TestJob\":8:{s:3:\"job\";N;s:10:\"connection\";N;s:5:\"queue\";N;s:15:\"chainConnection\";N;s:10:\"chainQueue\";N;s:5:\"delay\";N;s:10:\"middleware\";a:0:{}s:7:\"chained\";a:0:{}}"
  },
  "id": "8HCJIvi8qSCG7jOdDPelkj21p55pJEjb"
}
```

etudor:
I mean a raw json message that is being published by a nodejs backend to Rabbitmq:
{"attr":"value"}

adm-bome:
The consumers/workers can recreate the TestJob with parameters and handle/process the TestJob.

If your messages in RabbitMQ is not using this format (in the payload) you have to write your own consumer/worker.

What I would do is read the laravel code `Illuminate\Queue\Worker` class
you can read how the payload is made into a class. Or you can just make a worker with something as:

```php
$connectionName = 'rabbitmq';
[...]

PR #353, opened 2020-08-21
add factory create job

gpibarra:
inspired by [https://github.com/vyuldashev/laravel-queue-rabbitmq/issues/338](https://github.com/vyuldashev/laravel-queue-rabbitmq/issues/338)
I need to consume messages that do NOT respect Laravel Job format (messages that are produced by other Non-Laravel systems).
With the factory class you can manipulate the original raw message, to overwrite fields and delegate to a function that to process them.

*Example:*
app\Providers\AppServiceProvider.php
```
    public function register()
    {
        $this->app->bind(RabbitMQJobFactory::class, MyRabbitMQJobFactory::class);
    }
```
app\Classes\Queue\RabbitMq\MyRabbitMQJobFactory.php
```
<?php

namespace App\Classes\Queue\RabbitMq;


use Illuminate\Container\Container;
use VladimirYuldashev\LaravelQueueRabbitMQ\Queue\RabbitMQQueue;
use PhpAmqpLib\Message\AMQPMessage;
use VladimirYuldashev\LaravelQueueRabbitMQ\Queue\Jobs\RabbitMQJob;
use VladimirYuldashev\LaravelQueueRabbitMQ\Queue\Jobs\RabbitMQJobFactoryInterface;

class MyRabbitMQJobFactory implements RabbitMQJobFactoryInterface
{
    public function create(
        Container $container,
        RabbitMQQueue $rabbitmq,
        AMQPMessage $message,
        string $connectionName,
        string $queue
    )
    {
        // Adapt original message to format Laravel
        $originalBody = $message->getBody();
        $newBody = $this->getNewBody($originalBody);
        $message->setBody($newBody);

        return new RabbitMQJob(
            $container,
            $rabbitmq,
            $message,
            $connectionName,
            $queue
        );
    }

    /**
     * @param  string $body
     * @return string
     */
    protected function getNewBody(string $body) {
        $bodyArray = json_decode($body, true);
        if ($bodyArray === null) {
            return $body;
        }
        $newBodyArray = $this->getNewBodyArray($bodyArray);
        return json_encode($newBodyArray);
    }

    /**
     * @param  array $body
     * @return array
     */
    protected function getNewBodyArray(array $body) {
        $newBody = [
            "uuid" => null,
            "displayName" => "RabbitMqConsumer",
            "job" => "App\\Classes\\Queue\\RabbitMq\\RabbitMqConsumer@handle",
            "maxTries" => null,
            "maxExceptions" => null,
            "delay" => null,
            "timeout" => null,
            "timeoutAt" => null,
            "data" => [],
            "id" => null,
        ];
        $newBody['data'] = $body;
        return $newBody;
    }

}
```

app\Classes\Queue\RabbitMq\RabbitMqConsumer.php
```
<?php

namespace App\Classes\Queue\RabbitMq;

use Illuminate\Support\Arr;
use VladimirYuldashev\LaravelQueueRabbitMQ\Queue\Jobs\RabbitMQJob;

class RabbitMqConsumer
{

    /**
     * Execute the job.
     *
     * @return void
     */
    public function handle(RabbitMQJob $job)
    {
        $data = Arr::get($job->payload(), 'data');
        dump($data);
        $job->delete();
    }
}
```


MorrisonHotel:
@gpibarra can you fix code style pls?

gpibarra:
I cannot see the detail of the errors,
"Analysis was misconfigured
Configuration problem
The 'simplified_null_return' solver cannot be disabled because it is not enabled by its preset. "

MorrisonHotel:
@vyuldashev can you watch this PR pls? We definitely need something like this.

src/Consumer.php
```diff
@@ -84,7 +85,8 @@ public function daemon($connectionName, $queue, WorkerOptions $options): void
             function (AMQPMessage $message) use ($connection, $options, $connectionName, $queue): void {
                 $this->gotJob = true;
 
-                $job = new RabbitMQJob(
+                /** @var RabbitMQJob $job */
+                $job = app(RabbitMQJobFactory::class)->create(
```
vyuldashev, line 89: Considering performance, I would suggest resolving factory just once before calling basic_consume, instead of resolving it every time message is consumed.

src/LaravelQueueRabbitMQServiceProvider.php
```diff
@@ -43,6 +44,8 @@ public function register(): void
                 );
             });
 
+            $this->app->bind(RabbitMQJobFactory::class);
```
vyuldashev, line 47: What do you think of making this configurable through connection config?

src/Queue/RabbitMQQueue.php
```diff
@@ -230,7 +231,7 @@ public function pop($queue = null)
 
             /** @var AMQPMessage|null $message */
             if ($message = $this->channel->basic_get($queue)) {
-                return $this->currentJob = new RabbitMQJob(
+                return app(RabbitMQJobFactory::class)->create(
```
vyuldashev, line 234: $this->currentJob still should be set

src/LaravelQueueRabbitMQServiceProvider.php
gpibarra: Both alternatives are possible. In case of using the configuration connection, ´RabbitMQJobFactory.php´ would not be necessary and all this PR does not make sense. What you prefer is ok.
sorry, my english is very bad,

vyuldashev: My idea is to add `job_factory` config option with default value as `VladimirYuldashev\LaravelQueueRabbitMQ\Queue\Jobs\RabbitMQJobFactory::class`

gpibarra: I think it's great. Will you do it? or do you want me to do a new PR?

vyuldashev: I think you can do it in this PR.

gpibarra:
```
        'rabbitmq' => [

            'driver' => 'rabbitmq',
            'queue' => env('RABBITMQ_QUEUE', 'default'),
            'connection' => PhpAmqpLib\Connection\AMQPLazyConnection::class,

            'hosts' => [
                [
                    'host' => env('RABBITMQ_HOST', '127.0.0.1'),
                    'port' => env('RABBITMQ_PORT', 5672),
                    'user' => env('RABBITMQ_USER', 'guest'),
                    'password' => env('RABBITMQ_PASSWORD', 'guest'),
                    'vhost' => env('RABBITMQ_VHOST', '/'),
                ],
            ],

            'options' => [
                'ssl_options' => [
                    'cafile' => env('RABBITMQ_SSL_CAFILE', null),
                    'local_cert' => env('RABBITMQ_SSL_LOCALCERT', null),
                    'local_key' => env('RABBITMQ_SSL_LOCALKEY', null),
                    'verify_peer' => env('RABBITMQ_SSL_VERIFY_PEER', true),
                    'passphrase' => env('RABBITMQ_SSL_PASSPHRASE', null),
                ],
                'queue' => [
                    'job_factory' => \VladimirYuldashev\LaravelQueueRabbitMQ\Queue\Jobs\RabbitMQJobFactory::class,
                ],
            ],

            /*
             * Set to "horizon" if you wish to use Laravel Horizon.
             */
            'worker' => env('RABBITMQ_WORKER', 'default'),

         ],
```

src/Consumer.php
gpibarra: revised

gpibarra:
@vyuldashev Did you get to see this pr?

src/LaravelQueueRabbitMQServiceProvider.php
adm-bome: Can i ask why you would prefer an factory in this case?
I think its over complicated.

When people want to do something else than the default `RabitMQJob::class`, they have to create multiple classes to extend the functionality.

I have an other perspective on this mather....

In the config just write
```
'options' => [
    ...
    
    'queue' => [
        ...

        'job' => \VladimirYuldashev\LaravelQueueRabbitMQ\Queue\Jobs\RabbitMQJob::class,
        
        ...
        ],
            
        ...
    ],

```

In the case of messages not created by laravel, and won't respect the payload... 

just write your own RabbitMQJob class. 

```php
<?php

namespace App\Queue\Jobs;

use VladimirYuldashev\LaravelQueueRabbitMQ\Queue\Jobs\RabbitMQJob as BaseJob;

class RabbitMQJob extends BaseJob
{

    /**
     * Fire the job.
     *
     * @return void
     */
    public function fire()
    {
        $payload = $this->payload();

        $class = WhatheverClassName::class;
        $method = 'handle';

        ($this->instance = $this->resolve($class))->{$method}($this, $payload);
    }
}

`````
What i would do is extend the current RabbitMQJob::class and overwrite the 'fire()' method.
The original fire method expects an payload with a key 'job' and it's value has to be an class that exists.
So by overwriting this method you can do watherver you want with custom payloads.

gpibarra: Perfect!. We wait for the mergeo of PR [https://github.com/vyuldashev/laravel-queue-rabbitmq/pull/370](https://github.com/vyuldashev/laravel-queue-rabbitmq/pull/370)

gpibarra:
PR [https://github.com/vyuldashev/laravel-queue-rabbitmq/pull/370](https://github.com/vyuldashev/laravel-queue-rabbitmq/pull/370)

PR #355, opened 2020-08-25
Fix Worker is getting killed by timeout when no more jobs available

### LINKED ISSUE #352: [10.x] Queue worker is getting stopped by timeout when no new jobs available
- Laravel/Lumen version: 6.x
- RabbitMQ version: doesn't matter
- Package version: 10.2.2

**Describe the bug**

When running worker with --timeout argument worker is getting stopped after processing job when no new jobs available.

**Steps To Reproduce**

Just run the queue worker with --timeout argument, for example 10.
Send any job to worker and wait until the job will be processed.
After 10 seconds worker will be terminated by timeout.

**Additional text**
The reason of behavior I wrote above is this line of code:
https://github.com/vyuldashev/laravel-queue-rabbitmq/blob/master/src/Consumer.php#L96

As you can see in that line of code timeout handler is registered, but timeout handler is not cleaned up after the job is processed.
Original Laravel worker does it - https://github.com/illuminate/queue/blob/master/Worker.php#L164

codercms:
Fixes #352

### DIFF
--- a/src/Consumer.php
+++ b/src/Consumer.php
@@ -97,8 +97,7 @@ class Consumer extends Worker
                 }
 
                 $this->runJob($job, $connectionName, $options);
-                
-                
+
                 if ($this->supportsAsyncSignals()) {
                     $this->resetTimeoutHandler();
                 }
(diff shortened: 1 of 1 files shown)


adm-bome (approved):

khepin:
Can this be brought in to v10 as well or is that version not maintained anymore?

Issue #368, opened 2020-10-05
How to Consume Raw Message


filipbekic01:
Let's say i read messages from some queue which is filled by non-laravel application. Therefore, payload of message is not formated as Laravel expects. It would be awesome to have some kind of routing where i can specify which job should listen which queue. 

Is that possible right now maybe? 

This is required since in most cases you use RabbitMQ to communicate with another software.

adm-bome:
Hi,

Ofcourse...
But not with the laravel QueueAPI. Unless you can alter the payload before entering rabbitMQ queue.

ex.
by altering the other software to add an extra key in the message payload. This key is called 'job' and have to contain an existing class/job within your laravel app. That class gets called when resolved with some parameters and you can do whatever you want from there out.

An other option when your not able to alter the payload from your other software is by making an bridge between the to software packages.

1. Within your laravel app (or an separate service/app), create your own worker. 
2. With that worker read the raw messages from rabbitMQ. (this can be done by this library)
3. When you get an message, create a new Class and execute a method with the message payload you received.

Optionaly: 
5. Create a Laravel Job and set all it's properties.
5. Dispatch this newly created Job to an other queue within rabbitMQ (or an other messagebroker).
4. Start an laravel worker to read that second queue.

For reference: #338
Also see #353, this is a work in progress and makes this issue simpler to handle.

adm-bome:
I created some branches and PR's and i avoided complexity by avoiding factories and extra interfaces.
also thanks to @gpibarra for the afford in PR #353 

With this change handling non formatted laravel messages, becomes much easier.
You won't have to write your own worker or alter the payload upfront.

That said:
What i would do is create an Job and Worker to handle the Job ;) 
Also explained as transforming an Non Laravel message to an Laravel Message, so you can keep track of your progress with tools like Horizon.

adm-bome:
Closing issue due to inactivity and because there are several ways to handle raw messages. This lib now also supports an easier way to handle these cases.

PR #371, opened 2020-10-11
Feature/use custom job class v10

adm-bome:
An different way for people who have issues working with messages not created by the laravel QueueAPI
refference issues: #338, #368

and my vision on PR #353 without factory

### DIFF
--- a/.styleci.yml
+++ b/.styleci.yml
@@ -1,4 +1,1 @@
 preset: laravel
-
-disabled:
-  - simplified_null_return
--- a/README.md
+++ b/README.md
@@ -160,7 +160,87 @@ When you want to instruct RabbitMQ to reroute failed messages to a exchange or a
     // ...    
 ],
 ```
+### Use your own RabbitMQJob class
+Sometimes you have to work with messages published by another application.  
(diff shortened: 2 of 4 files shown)


adm-bome (commented):
@vyuldashev, we can close the PR for the master. If you can cherrypick or merge V10 branch into the master branch?

vyuldashev:
@adm-bome According to our support policy we add new features only to the latest version.

adm-bome:
Yep, i know.

This feature was introduced at the same time branch version 11 was released.

Version 11 was created only for  the new horizon and is not backwardcompattible, because of that.

That meens versions of laravel 6 and 7 with an old structure can't use this feature.

As i see it... when the feature is also in version 10 more people can use this. (Including myself) 

The changes don't interfer with support for horizon 3 as of version 11.

adm-bome:
Ps: by eccident i restored the branch. 

Maybe you can reconsider your choice of not implementing this feature into v10. 
Its not a major change for how the library is working.

But i will respect your choice when you think its not worth it and please delete the branch

PR #385, opened 2020-12-09
add support for defining after creating connection callbacks

vyuldashev:
This PR allows to define custom callbacks to run after connection is created:

For example, in AppServiceProvider:

```php
use PhpAmqpLib\Connection\AbstractConnection;
use PhpAmqpLib\Connection\Heartbeat\PCNTLHeartbeatSender;
use VladimirYuldashev\LaravelQueueRabbitMQ\Queue\Connectors\RabbitMQConnector;

RabbitMQConnector::afterCreatingConnection(function(AbstractConnection $connection) {
    $sender = new PCNTLHeartbeatSender($connection);
    $sender->register();
});
```

### DIFF
--- a/src/Queue/Connectors/RabbitMQConnector.php
+++ b/src/Queue/Connectors/RabbitMQConnector.php
@@ -2,6 +2,7 @@
 
 namespace VladimirYuldashev\LaravelQueueRabbitMQ\Queue\Connectors;
 
+use Closure;
 use Exception;
 use Illuminate\Contracts\Events\Dispatcher;
 use Illuminate\Contracts\Queue\Queue;
@@ -21,7 +22,9 @@ class RabbitMQConnector implements ConnectorInterface
     /**
      * @var Dispatcher
      */
-    private $dispatcher;
+    protected $dispatcher;
+
+    protected static $afterConnectionCreatingCallbacks = [];
 
     public function __construct(Dispatcher $dispatcher)
     {
@@ -39,6 +42,8 @@ class RabbitMQConnector implements ConnectorInterface
     public function connect(array $config): Queue
     {
         $connection = $this->createConnection(Arr::except($config, 'options.queue'));
+
+        $this->callAfterCreatingConnection($connection);
 
         $queue = $this->createQueue(
             Arr::get($config, 'worker', 'default'),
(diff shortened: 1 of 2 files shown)


adm-bome (commented):
I would add an extra parameter for adding callbacks to a specific connection some how. So every connection can have different callbacks or none.
Maybe work with a Callable instead if a closure so some one can add an class or a closure
Maybe use the containers logic to call the callable with container parameter.

A complete diffrent approach is using the event system. On every connection in config, you can set an option 'useConnectionCallbacks' to publish the AfterConnectionCreated event with the connection as an variable. 

Then in the eventServiceProvider define your callbacks

vyuldashev:
@adm-bome Extra parameter won't help because `connect` method accept configuration without connection name. But, we could do something like `$callback($connection, $config);`, so there would be a possibility to add extra keys to configuration in order to be able to distinguish connections, if someone will need it.

adm-bome:
> Extra parameter won't help because `connect` method accept configuration without connection name.

Yes i just saw the interface...

I edited my review above. I would go for an approach of the event system.

vyuldashev:
@adm-bome Good idea with setting this in configuration. We could simplify that to:

```php
return [
    'driver' => 'rabbitmq',

    'callbacks' => [
        'after_creating_connection' => [
            AfterCreatingConnectionCallback::class,
            function(\PhpAmqpLib\Connection\AbstractConnection $connection) {
    
            },
        ],
    ],
];
```

vyuldashev:
@adm-bome Or:

```php
// config.php
return [
    'driver' => 'rabbitmq',

    'middleware' => DefaultConnectionMiddleware::class,
];

class DefaultConnectionMiddleware implements RabbitMQMiddleware
{
    public function connectionCreated(): void
    {
        // ...
    }
}
```

adm-bome:
@vyuldashev 
Hahaha, so many options to choose from. But the middle ware strategy is more for when you want to change, tranform or interact on the connection. And feels like overkill.

When i think about it, the event system is the easiest to implement.
The dispacher is already added via dependicies 

So creating an eventClass, extra option in config and the check to dispatch is just a couple of lines.

vyuldashev:
@adm-bome Could you, please, provide an example from developer's perspective?

adm-bome:
Tomorrow I will, its a little hard on my phone 😉

Issue #384, opened 2020-12-09
How to use PCNTLHeartbeatSender with this?


fredsted:
I'm wondering how I should implement `PCNTLHeartbeatSender` with this package. We use this package to run some very long-running jobs (30 minutes +), and we seem to get `AMQPChannelClosedException` exceptions after a while.

Due to PHP's single threaded nature, it seems that the only way to run long-running tasks using RabbitMQ this is to use php-amqplib's new PCNTLHeartbeatSender feature, as documented here: https://github.com/php-amqplib/php-amqplib/pull/815

I would like to use this in my own project, but I'm not sure where I should put the `new PCNTLHeartbeatSender` code. Does anyone have a pointer on this? Thank you.

vyuldashev:
@fredsted Hello! Take a look at #385, please. What do you think?

fredsted:
@vyuldashev Thanks so much for the quick response. It looks good to me. I tried this method and I got a segmentation fault a couple of seconds after I ran the `queue:work` console command.

It seems that the only way to make PCNTLHeartbeatSender work reliably is to use it in the job class, e.g.:

```
/** @property RabbitMQJob $job */
class SleepTest implements ShouldQueue
{
    use InteractsWithQueue;

    public function handle(): void
    {
        $sender = new PCNTLHeartbeatSender($this->job->getRabbitMQ()->getConnection());
        $sender->register();

        dump('before_sleep');
        $sleeps = 0;
        while ($sleeps < 20) {
            $sleeps++;
            dump('sleep '.$sleeps);
            sleep(1);
        }
        dump('after_sleep');

        $sender->unregister();
    }
}
```

If I initialize it somewhere else than the job class, I either get a segfault or it just doesn't work. For reference, I also tried a custom Connection class which I've used in the `queue.connections.rabbitmq.connection` config value, for example:

```
class AmqpConnection extends AMQPSocketConnection
{
    private static PCNTLHeartbeatSender $sender;

    public static function create_connection($hosts, $options = array())
    {
        $connection = parent::create_connection($hosts, $options);

        if ($connection->isConnected()) {
            self::$sender = new PCNTLHeartbeatSender($connection);
            self::$sender->register();
            dump('registered');
        }

        return $connection;
    }
}
```

[...]

PR #386, opened 2020-12-10
Feature/abbility to hook into the created connection

adm-bome:
@vyuldashev What do you think?
Issue: #384 PR: #385

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,6 +250,72 @@ class RabbitMQJob extends BaseJob
     }
 }
 ```
+### PhpAmqpLib Connection - Created event (hook)
+For every defined rabbitmq connection in the `queue.connections` config, a connection is created. 
+After the connection was successfully created an `ConnectionCreated` Event is fired, with the parameters `connection` and `tags`.
+
+Tagging a connection in the config via the property `connection_tags`, makes it possible to run your Listeners only for specific connections.   
+When not tagging your connection there is no way of knowing which connection was created or manage your listeners in executing. 
+
+An example for sending a heartbeat signal against the created connection:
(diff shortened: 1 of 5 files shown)


adm-bome:
@vyuldashev i hope this feature is implemented.
Why you may ask?
The reasons are:

1. @fredsted is coming with a question about a functionality the library Php-AmqpLib is providing. It's a matter of time someone else asks the same question.
2. Some-one does not have to use this feature when they don't need it. Laravel and all other libraries/packages are filled with unused events.
3. Another thing is that the middleware approach is ok-isch and works on a per job basis, its alot of overhead for every job. When you want to register the PCNTLHeartbeatSender on a connection basis this is the approach everyone should use.
4. This feature is an hook. It does not have to mean, a programmer uses this event for just the PCNTLHeartbeatSender.

This feature makes the lib more flexible, and easier to work with over-all.
>Side note: We don't have to implement this feature right now. we could put it aside and on the road-map for the next release

vyuldashev:
@adm-bome Let's close it until someone will need such functionality. I don't want to add code which is not linked to needs.

frops:
We are need to implement this feature, because our worker not listening after some times. I have read https://blog.mollie.com/keeping-rabbitmq-connections-alive-in-php-b11cb657d5fb and I need to implementation.

Another way - is inheritance  LaravelQueueRabbitMQServiceProvider и RabbitMQConnector, but it's not good way (
Idea with events is very good.
Thank you ;)

frops:
@adm-bome @vyuldashev Can you reopen pull request?

jvanschie:
I also need this feature

PR #404, opened 2021-02-02
Apply timeout reset fix (#355) to v10.0 branch

khepin:
The fix in https://github.com/vyuldashev/laravel-queue-rabbitmq/pull/355 has made it to v11 but is still an issue on v10.


szainmehdi (approved):
👍

Issue #408, opened 2021-02-24
Worker ignores `illuminate:queue:restart` signal


frans-beech-it:
- Laravel version: 8.26.1
- RabbitMQ version: 3
- Package version: 11.1.1

**Describe the bug**

After calling `queue:restart` the worker doesn't stop. 

**Steps To Reproduce**

1. Run `php artisan rabbitmq:consume`
2. Run in different terminal `php artisan queue:restart`

**Current behavior**

Worker ignores restart signal

**Expected behavior**

Worker should stop when `stopIfNecessary()` returns a exit code

**Additional context**

on https://github.com/vyuldashev/laravel-queue-rabbitmq/blob/master/src/Consumer.php#L142 now the `stopIfNecessary()` is called. But the exit code isn't handled. It should break from the `while` loop and stop the "deamon"

```php
$this->stopIfNecessary($options, $lastRestart, $this->gotJob ? true : null);
```

should be changed to:

```php
            $status = $this->stopIfNecessary(
                $options, $lastRestart, $startTime, $jobsProcessed, $job
            );

            if (! is_null($status)) {
                return $this->stop($status);
            }
```

MorrisonHotel:
@vyuldashev same problem.
We use systemd for workers, and systemd just killing the process with timeout exception on trying to restart service.

khepin:
Version 11 of this package is no longer supported. If you are still experiencing the issue with a more recent version, please open a separate issue.

PR #409, opened 2021-02-24
Update Consumer to stop when stopIfNecessary() returns exit code

frans-beech-it:
#408

### DIFF
--- a/src/Consumer.php
+++ b/src/Consumer.php
@@ -54,7 +54,15 @@ class Consumer extends Worker
         $this->prefetchCount = $value;
     }
 
-    public function daemon($connectionName, $queue, WorkerOptions $options): void
+    /**
+     * Listen to the given queue in a loop.
+     *
+     * @param  string  $connectionName
+     * @param  string  $queue
+     * @param  \Illuminate\Queue\WorkerOptions  $options
+     * @return int
+     */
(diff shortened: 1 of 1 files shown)


src/Consumer.php
```diff
@@ -139,7 +139,13 @@ function (AMQPMessage $message) use ($connection, $options, $connectionName, $qu
             // Finally, we will check to see if we have exceeded our memory limits or if
             // the queue should restart based on other indications. If so, we'll stop
             // this worker and let whatever is "monitoring" it restart the process.
-            $this->stopIfNecessary($options, $lastRestart, $this->gotJob ? true : null);
+            $status = $this->stopIfNecessary(
+                $options, $lastRestart, $startTime, $jobsProcessed, $job
+            );
+
+            if (! is_null($status)) {
+                return $this->stop($status);
```
MorrisonHotel, line 147: A void function must not return a value

frans-beech-it: ah, return is invalid. See parent class from laravel https://github.com/laravel/framework/blob/bc7d4b7fb1a444c1646d7d47a75185b72e78537d/src/Illuminate/Queue/Worker.php#L117

vyuldashev:
Released under v11.1.2. Thanks!

frans-beech-it:
:+1:

PR #422, opened 2021-05-17
Add priority for jobs

sujit-baniya:
Sometimes it is essential for a job to be consumed based on priority. So the changes allow users to set priority on Consumer and Jobs.

### DIFF
--- a/src/Console/ConsumeCommand.php
+++ b/src/Console/ConsumeCommand.php
@@ -25,7 +25,7 @@ class ConsumeCommand extends WorkCommand
                             {--tries=1 : Number of times to attempt a job before logging it failed}
                             {--rest=0 : Number of seconds to rest between jobs}
 
-                            {--max-priority=4}
+                            {--max-priority}
                             {--consumer-tag}
                             {--prefetch-size=0}
                             {--prefetch-count=1000}
--- a/src/Console/ConsumeCommand.php
+++ b/src/Console/ConsumeCommand.php
@@ -24,7 +24,8 @@ class ConsumeCommand extends WorkCommand
                             {--timeout=60 : The number of seconds a child process can run}
                             {--tries=1 : Number of times to attempt a job before logging it failed}
                             {--rest=0 : Number of seconds to rest between jobs}
-                           
+
+                            {--max-priority=4}
                             {--consumer-tag}
                             {--prefetch-size=0}
                             {--prefetch-count=1000}
@@ -40,6 +41,7 @@ class ConsumeCommand extends WorkCommand
         $consumer->setContainer($this->laravel);
         $consumer->setName($this->option('name'));
         $consumer->setConsumerTag($this->consumerTag());
+        $consumer->setMaxPriority((int) $this->option('max-priority'));
         $consumer->setPrefetchSize((int) $this->option('prefetch-size'));
         $consumer->setPrefetchCount((int) $this->option('prefetch-count'));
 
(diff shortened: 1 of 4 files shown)


src/Console/ConsumeCommand.php
```diff
@@ -24,7 +24,8 @@ class ConsumeCommand extends WorkCommand
                             {--timeout=60 : The number of seconds a child process can run}
                             {--tries=1 : Number of times to attempt a job before logging it failed}
                             {--rest=0 : Number of seconds to rest between jobs}
-                           
+
+                            {--max-priority=4}
```
vyuldashev, line 28: Why 4?

src/Consumer.php
```diff
@@ -42,6 +45,11 @@ public function setConsumerTag(string $value): void
         $this->consumerTag = $value;
     }
 
+    public function setMaxPriority(int $value): void
+    {
+        $this->maxPriority = $value ?? 1;
```
vyuldashev, line 50: value is not nullable, so `?? 1` is redundant

sujit-baniya: Fixed

src/Console/ConsumeCommand.php
sujit-baniya: IT was added as 4 as I thought it would describe as

1 - Low
2 - Normal
3 - High
4 - Critical

vyuldashev: I guess the queue should be declared with `x-max-priority` in order to support this feature?

sujit-baniya: @vyuldashev The queue `x-max-priority` is obtained from config as mentioned in

RabbitMQQueue.php

```php
    protected function getQueueMaxPriority(): int
    {
        return (int) (Arr::get($this->options, 'queue_max_priority') ?: 2);
    }
 ```
 Maybe that should suffice?

When a consumer runs on a queue with `php artisan rabbitmq:consume` the new queue will have `x-max-priority` set.

For the QueueDeclareCommand, Do you think it requires to add option for `max-priority` ?

Comment (sujit-baniya) [reply src/Console/ConsumeCommand.php:28]: @vyuldashev I've added the change for QueueDeclare as well. Please have a look on commit
https://github.com/vyuldashev/laravel-queue-rabbitmq/pull/422/commits/9cd95be5414873c8f861f301da40b84b2be95522

Comment (sujit-baniya) [reply src/Console/ConsumeCommand.php:28]: @vyuldashev When I tested priorities on local, I found setting `prefetch-count=1000` (high prefetch count) is not wise as priorities are not considered at all for jobs within those 1000. So maybe need to decrease default to 5 or any number below two digits

Comment (vyuldashev) [reply src/Console/ConsumeCommand.php:28]: @sujit-baniya According to the [docs](https://www.rabbitmq.com/priority.html):

```
The AMQP 0-9-1 spec is a bit vague about how priorities are expected to work. It says that all queues MUST support at least 2 priorities, and MAY support up to 10. It does not define how messages without a priority property are treated.

In contrast to the AMQP 0-9-1 spec, RabbitMQ queues by default do not support priorities. When creating priority queues, a maximum priority can be chosen as developer sees fit. When choosing the value, a couple of things must be taken into account.

There is some in-memory and on-disk cost per priority level per queue. There is also an additional CPU cost, especially when consuming, so you may not wish to create huge numbers of levels.

The message priority field is defined as an unsigned byte, so in practice priorities should be between 0 and 255.

Messages without a priority property are treated as if their priority were 0. Messages with a priority which is higher than the queue's maximum are treated as if they were published with the maximum priority.
```

So, I think priority queues should be an option, not a default feature. Can you adjust your code to use priority only when a dev is willing to, please?

sujit-baniya: @vyuldashev  Yes I've updated the code to use priority as option

sujit-baniya:
@vyuldashev I've added the priority queue as optional

PR #428, opened 2021-06-27
RPC support by injecting correlation_id & reply_to from options

khalilst:
In the following of support for **RPC** (Remote Procedure Call), this PR allows to use `correlation_id` from `options` argument in the `pushRaw` method.
Also, it provides to inject the `reply_to` property using options.

I have followed the instructions from [RabbitMQ RPC documentation](https://www.rabbitmq.com/tutorials/tutorial-six-javascript.html).


amiranbarigreenweb:
Hi, how can we send obejct of a job for payload? @khalilst

PR #447, opened 2021-11-12
Add option to enable GZIP'ing messages

khepin:
Hi, we've got a queue worker that generates fairly large messages that contain a batch of documents to be indexed in elasticsearch.

The size of the messages makes rabbitmq really unstable when we start using this due to the high memory usage this causes.

While it seems like we'll be able to move along with this by using the `worker` and `job` config options, this seemed like it could be valuable enough to have in this library itself.

Let me know if this goes in the right direction / if there are things you'd like to be updated etc...

### DIFF
--- a/src/Queue/Jobs/RabbitMQJob.php
+++ b/src/Queue/Jobs/RabbitMQJob.php
@@ -36,6 +36,13 @@ class RabbitMQJob extends Job implements JobContract
      */
     protected $decoded;
 
+    /**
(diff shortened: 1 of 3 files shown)


khepin:
Locally `composer test:style` or `composer fix:style` doesn't report any issue. Not sure why styleci is failing then

yuehlin-square (approved):

szainmehdi (approved):

PR #468, opened 2022-02-28
Dispatch a job after DB transaction commit 

shahramfdl:
Support Laravel 8 and later

- The specific job should be dispatched after all open database transactions have been committed
 - [https://laravel.com/docs/8.x/queues\#specifying-commit-dispatch-behavior-inline](https://laravel.com/docs/8.x/queues/#specifying-commit-dispatch-behavior-inline)

Example:
Using when you want to dispatch
`TestJob::dispatch()->afterCommit();`

or you can use it on your job
```
class TestQueue implements ShouldQueue
{
    use Dispatchable, InteractsWithQueue, Queueable, SerializesModels;

    public function __construct()
    {
        $this->onConnection('rabbitmq');
        $this->onQueue('test');
        $this->afterCommit();
    }

    public function handle()
    {
        //...
    }
}
```

also it simply use on event listener queueable
```
class TestListener implements ShouldQueue
{
    use InteractsWithQueue;

    public $afterCommit = true;

    public function viaConnection()
    {
        return 'rabbitmq';
    }

    public function handle($event)
    {
        //...
    }
}
```

### DIFF
--- a/src/Queue/Connectors/RabbitMQConnector.php
+++ b/src/Queue/Connectors/RabbitMQConnector.php
@@ -44,6 +44,7 @@ class RabbitMQConnector implements ConnectorInterface
             Arr::get($config, 'worker', 'default'),
             $connection,
             $config['queue'],
+            $config['after_commit'] ?? null,
             Arr::get($config, 'options.queue', [])
         );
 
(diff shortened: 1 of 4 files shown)


jehoshua02:
@shahramfdl will this also support `after_commit` from `config/queue.php`? Is that coming in a later PR?

shahramfdl:
> @shahramfdl will this also support `after_commit` from `config/queue.php`? Is that coming in a later PR?

Enable "after_commit" wherever Laravel allows you, yes its work now.

Example of specific RabbitMQ connection:

```
'rabbitmq' => [

            'driver' => 'rabbitmq',
            'queue' => env('RABBITMQ_QUEUE', 'default'),
            'connection' => PhpAmqpLib\Connection\AMQPLazyConnection::class,
            'after_commit' => true,
            ....
],
```
It works like a charm.

judgej:
Running Laravel 9 and using this driver neither the `->afterCommit()` method nor the `'after_commit' => true` config option were working. That left if scratching our heads for some time - stuff would work using `redis` queues then fail with models not found using this rabbitmq driver.

This PR fixed it immediately. There are a number of other PRs and fixes on branches not submitted as PRs, that seem to be fixing things such as the `->delay()` option, which is a little concerning. Without them, this is a broken driver. What needs to happen to get these PRs looked at?

vyuldashev (approved):

vyuldashev:
Sorry for the long delay. I appreciate the work. Since this PR introduces breaking changes to the API, I release it under v13.0.0

judgej:
Cool :-) We are going to be running some high load tests tomorrow with this PR included, so can let you know how it goes, if that helps at all. Much appreciated.

Issue #318, opened 2020-03-03
[Feature] Work with multiple queue on rabbitmq:consumer command


itsursujit:
- Lumen Version:6.3
- Package Version: latest

When running `php artisan rabbitmq:command --queue="abc,default,xyz"` I'm getting following error

```
In AMQPChannel.php line 215:
                                                                               
  NOT_FOUND - no queue 'store-to-db-queue,store-to-db-prio-3,store-to-db-prio  
  -2,store-to-db-prio-1,store-to-db-prio-0' in vhost '/'  
```
As a possible solution I tried to do following:

```
    $queueName = explode(',', $queue);
        foreach ($queueName as $name) {
            $name = $this->mode . $name;
            $this->channel->queue_declare(
                $name,
                false,
                true,
                false,
                false,
                false
            );
            $this->channel->basic_consume(
                $name,
                '',
                false,
                false,
                false,
                false,
                function (AMQPMessage $message) use ($connection, $options, $connectionName, $name): void {
                    $this->gotJob = true;

                    $job = new RabbitMQJob(
                        $this->container,
                        $connection,
                        $message,
                        $connectionName,
                        $name
                    );

                    if ($this->supportsAsyncSignals()) {
                        $this->registerTimeoutHandler($job, $options);
                    }

                    $this->runJob($job, $connectionName, $options);
                }
            );
        }
```
With above change, jobs are working but when trying to Release a job or retry a job or delay a job,
I'm getting following error:

`PRECONDITION_FAILED - unknown delivery tag 3`

How to fix such issue?

adm-bome:
Try the master until the new code is released.
See also my [comment](https://github.com/vyuldashev/laravel-queue-rabbitmq/pull/303/files#r375932387)

`composer require vyuldashev/laravel-queue-rabbitmq:dev-master`

> **Side note**: 
As long there are no massages in the queue(s) or the queue(s) do not exist in rabbitMq, the worker(s)/consumer(s) should wait or time-out. In _most_ cases it's not the responsibility of a worker/consumer to create a queue. In most cases this is the responsibility of RabbitMq or the producer of the message.

itsursujit:
@adm-bome The issue is I cannot run a consumer with multiple queue:
`php artisan rabbitmq:consume  --queue=store-to-db-queue,store-to-db-prio-3,store-to-db-prio-2,store-to-db-prio-1,store-to-db-prio-0`

adm-bome:
[...]

PR #470, opened 2022-03-10
Document missing support for multiple queues 

mortenhauberg:
While trying to figure out why I couldn't consume from multiple queues when using `php artisan rabbitmq:consume` I found https://github.com/vyuldashev/laravel-queue-rabbitmq/issues/318#issuecomment-626195498, saying that there will be no support for this.

These changes will, hopefully, spare others from the archaeological work I had to do.


M-Porter:
Thanks for the callout @mortenhauberg. Do you think it is also worth adding that if you must consume multiple queues, that you should use the build-in laravel `php artisan queue:work` command?

mortenhauberg:
> Thanks for the callout @mortenhauberg. Do you think it is also worth adding that if you must consume multiple queues, that you should use the build-in laravel `php artisan queue:work` command?

Sure thing. I've added a note to the readme. Let me know if you want it phrased differently

PR #487, opened 2022-10-17
Show worker name in connections RabbitMQ admin management

i3bepb:
In console command exist options worker name. I suggest using this name in connections RabbitMQ admin management.
```sh
./artisan rabbitmq:consume --name="Consumer RaSmartId"
```

Now it looks like this:

Look bad

After change:

Look good

https://github.com/php-amqplib/php-amqplib/issues/728

### DIFF
--- a/src/Consumer.php
+++ b/src/Consumer.php
@@ -79,6 +79,9 @@ class Consumer extends Worker
 
         [$startTime, $jobsProcessed] = [hrtime(true) / 1e9, 0];
 
+        $phpAmqpLibConnection = $this->container['config']["queue.connections.$connectionName.connection"];
+        $phpAmqpLibConnection::$LIBRARY_PROPERTIES['connection_name'] = ['S', $options->name];
+
         /** @var RabbitMQQueue $connection */
         $connection = $this->manager->connection($connectionName);
 


everflux:
Is there a reason not to user the AMQPConnectionConfig?
For example

```
 $config = new AMQPConnectionConfig();
 $config->setConnectionName("PHP-worker-node-1");
    $connection = new AMQPSSLConnection( 'localhost', 1234, 'guest',   'guest', '/', $ssl_options, array(), 'ssl', $config);
```

i3bepb:
@everflux 
> Is there a reason not to user the AMQPConnectionConfig? For example

The method ```setConnectionName``` was added only in [php-amqplib/php-amqplib](https://github.com/php-amqplib/php-amqplib) 3.4, before it was not. I suggested changes before the release 3.4.

Now yes it is correct to use method ```setConnectionName```

khepin (approved):

src/Consumer.php
```diff
@@ -79,6 +79,9 @@ public function daemon($connectionName, $queue, WorkerOptions $options)
 
         [$startTime, $jobsProcessed] = [hrtime(true) / 1e9, 0];
 
+        $phpAmqpLibConnection = $this->container['config']["queue.connections.$connectionName.connection"];
```
M-Porter, line 82: I think we'd want to only set the name if this config value is set.

I'm not sure everyone would like to see `\PhpAmqpLib\Connection\AMQPLazyConnection` start appearing by default as the connection name in the UI.

M-Porter (changes requested):

M-Porter:
Closing as resolved by https://github.com/vyuldashev/laravel-queue-rabbitmq/pull/528

PR #496, opened 2022-11-09
Add $dispatchAfterCommit when running via custom worker

wietsewarendorff:


jvanschie:
@vyuldashev can you please look into this?

mattgrande:
Hi @wietsewarendorff, thanks for your submission. Unfortunately, we can't accept it as-is, since it would break any existing workers as soon as it's merged. We're evaluating other options (most likely a Service Provider) for use in the next major version.

khepin:
Duplicate of #513 , closing this one and moving the convo over there.

PR #513, opened 2023-01-26
pass dispatchAfterCommit into custom worker constructor

oprudkyi:
align constructors for custom workers and RabbitMQQueue

so there no need to map somehow constructor in case of custom worker 

```
'worker' => env('RABBITMQ_WORKER', App\Services\CustomRabbitMQQueue::class),

...

use VladimirYuldashev\LaravelQueueRabbitMQ\Queue\RabbitMQQueue;

class CustomRabbitMQQueue extends RabbitMQQueue
{
```

### DIFF
--- a/src/Queue/Connectors/RabbitMQConnector.php
+++ b/src/Queue/Connectors/RabbitMQConnector.php
@@ -106,7 +106,7 @@ class RabbitMQConnector implements ConnectorInterface
             case 'horizon':
                 return new HorizonRabbitMQQueue($connection, $queue, $dispatchAfterCommit, $options);
             default:
-                return new $worker($connection, $queue, $options);
+                return new $worker($connection, $queue, $dispatchAfterCommit, $options);
         }
     }
 


M-Porter:
I believe the intention for this was for custom workers to extend [Queue](https://github.com/laravel/framework/blob/205fae4a81945637dcd11628664aa96e00096e0a/src/Illuminate/Queue/Queue.php) and implement [QueueContract](https://github.com/laravel/framework/blob/205fae4a81945637dcd11628664aa96e00096e0a/src/Illuminate/Contracts/Queue/Queue.php).

That being said, it is a good idea to make those calls consistent.

khepin:
In its current form, this would be a breaking change and will therefore have to wait on a major version release.
If we wanted to allow a similar behavior prior to that, we'd need to offer a different way of achieving the same result.

oprudkyi:
@khepin what if there will be some new option
```
'worker' => env('RABBITMQ_WORKER', App\Services\CustomRabbitMQQueue::class),
'custom_worker_type' =>  'with_dispatch_after_commit',
...
this will allow to have fix without introduction of breaking changes

**Comment by adm-bome** [comment]
Does this work ????

- No library code adjustments, only adjustments in your own Class and config.


```php
<?php

namespace App\Queue;

use Illuminate\Support\Arr;
use VladimirYuldashev\LaravelQueueRabbitMQ\Queue\RabbitMQQueue;

class CustomRabbitMQQueue extends RabbitMQQueue
{

    public function __construct(
        AbstractConnection $connection,
        string $default,
        array $options = []
    ) {
        $dispatchAfterCommit = Arr::pull($options, 'after_commit', false);
        parent::__construct($connection, $default, $dispatchAfterCommit, $options);
    }

```
and in the connection config ( move the `after_commit` option to the queue options)

```php

#/config/queue.php
'connections' => [
    // ...

    'rabbitmq' => [
        // ...

        'options' => [
            // 'after_commit' => true
            'queue' => [
                // ...
                'after_commit' => true
                'job' => \App\Queue\Jobs\RabbitMQJob::class,
            ],
        ],
    ],

    // ... 
```

adm-bome:
@khepin 
And I agree to align the constructor calls. 

When  `dispatchAfterCommit` was introduced into Laravel and eventually was implemented into this library... the contructor calls had to be already aligned. This library checks if the Worker class (Queue) is an instance of RabibitMQQueue:class, so the call is wrong because the constructor of the RabibitMQQueue:class is already different.

If people are depending on the constructor of the RabibitMQQueue:class, people had already noticed issues and had to correct this with a sort of sollution I presented above.

oprudkyi:
Hi @adm-bome 
everything works (currently we're using fork, I think using own RabbitMQConnector is even more simpler option ) , but this PR will allow to avoid introducing yet another dependency in our code

adm-bome:
> I think using own RabbitMQConnector is even more simpler option

Not really, if you want to replace all classes then use a fork or write your own implementation of a driver :smile: 

The contructors calls must align with the default constructor of RabibitMQQueue:class.

As i mentioned above `_the contructor calls had to be already aligned_` and your problem should have never emerge.
When this was noticed in the beginning, when review-ing the initial merge of the feature `dispatchAfterCommit`, all was good for now.

PR #521, opened 2023-02-17
pass dispatchAfterCommit into custom worker constructor wihtout breaking changes

oprudkyi:
- align constructors for custom workers and RabbitMQQueue
- add extra option 'custom_worker_type' =>  'with_dispatch_after_commit', to allow change type of constructor (not required, default - old behavior)

this is version of https://github.com/vyuldashev/laravel-queue-rabbitmq/pull/513 without breaking changes (things handled via extra option)

@khepin please check

### DIFF
--- a/src/Queue/Connectors/RabbitMQConnector.php
+++ b/src/Queue/Connectors/RabbitMQConnector.php
@@ -42,6 +42,7 @@ class RabbitMQConnector implements ConnectorInterface
 
         $queue = $this->createQueue(
             Arr::get($config, 'worker', 'default'),
+            Arr::get($config, 'custom_worker_type', 'default'),
             $connection,
             $config['queue'],
             Arr::get($config, 'after_commit', false),
@@ -95,6 +96,7 @@ class RabbitMQConnector implements ConnectorInterface
      */
     protected function createQueue(
         string $worker,
+        string $customWorkerType,
         AbstractConnection $connection,
         string $queue,
         bool $dispatchAfterCommit,
@@ -106,7 +108,12 @@ class RabbitMQConnector implements ConnectorInterface
             case 'horizon':
                 return new HorizonRabbitMQQueue($connection, $queue, $dispatchAfterCommit, $options);
             default:
-                return new $worker($connection, $queue, $options);
+                switch ($customWorkerType) {
+                    case 'with_dispatch_after_commit':
+                        return new $worker($connection, $queue, $dispatchAfterCommit, $options);
+                    default:
+                        return new $worker($connection, $queue, $options);
+                }
         }
     }
 
--- a/README.md
+++ b/README.md
@@ -254,6 +254,62 @@ class RabbitMQJob extends BaseJob
 }
 ```
 
+### Use your own RabbitMQQueue class
+
+To use own RabbitMQQueue set `RABBITMQ_WORKER` (or `worker` param) to your queue class.
+
+```php
+RABBITMQ_WORKER='App\\Services\\CustomRabbitMQQueue'
+```
+or
+
+```php
+    'worker' => env('RABBITMQ_WORKER', App\Services\CustomRabbitMQQueue::class),
+```
+
+where queue class is defined like
+```php
+<?php
+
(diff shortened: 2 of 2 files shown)


src/Queue/Connectors/RabbitMQConnector.php
```diff
@@ -106,7 +108,12 @@ protected function createQueue(
             case 'horizon':
                 return new HorizonRabbitMQQueue($connection, $queue, $dispatchAfterCommit, $options);
             default:
-                return new $worker($connection, $queue, $options);
+                switch ($customWorkerType) {
+                    case 'with_dispatch_after_commit':
+                        return new $worker($connection, $queue, $dispatchAfterCommit, $options);
+                    default:
+                        return new $worker($connection, $queue, $options);
+                }
```
M-Porter, line 116: this feels like a hack rather than a solution to the problem.

would like to see a move towards using factory make functions that return a class so we don't have to rely on params being perfectly matched up as long as they all share an underlying contract. this is something @khepin and I have discussed but are yet to really dig into a good solution yet.

src/Queue/Connectors/RabbitMQConnector.php
```diff
@@ -95,6 +96,7 @@ protected function createConnection(array $config): AbstractConnection
      */
     protected function createQueue(
         string $worker,
+        string $customWorkerType,
```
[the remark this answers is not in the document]
oprudkyi: Hi @M-Porter 
I some kind of confused
- as for me this is a minor change (only internal api has changed, protected method) and this can be released as minor version 
- the library by self has recently released breaking changes as minor version , by example https://github.com/vyuldashev/laravel-queue-rabbitmq/releases/tag/v13.1.0

src/Queue/Connectors/RabbitMQConnector.php
oprudkyi: well, the library right now have a major bug which blocks real use of it (because call incompatible with Laravel )
I would prefer it fixed in some compatible way (right now we rely on custom fork)

adm-bome: nope, nope, nope

This does not only feels like a hack... it is a hack.

src/Queue/Connectors/RabbitMQConnector.php
```diff
@@ -42,6 +42,7 @@ public function connect(array $config): Queue
 
         $queue = $this->createQueue(
             Arr::get($config, 'worker', 'default'),
+            Arr::get($config, 'custom_worker_type', 'default'),
```
adm-bome, line 45: This config item does not follow other connection properties available to all drivers. 
Rather put this in the options property [ options => [ 'custom_option' => 'value' ]]

adm-bome (commented):
Won't approve such changes.


> This can be solved/addressed, without aletering the library's code.
 As I already wrote above.

src/Queue/Connectors/RabbitMQConnector.php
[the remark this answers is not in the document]
M-Porter: To be clear: v13.1.0 was a breaking minor version by the laravel library itself. We had to no choice but to follow suite as doing `composer update` for a patch version would have broken so to mitigate the problem created by laravel that was the only solution.

Yes, it's protected, but nothing keeps you from writing your own connector which others do and this would break `parent::createQueue()` calls.

src/Queue/Connectors/RabbitMQConnector.php
M-Porter: > the library right now have a major bug which blocks real use of it

Nothing is preventing you from creating you own `$worker` class compatible with the underlying Queue contracts and in the constructor doing what you need to. You do not **have** to extend `VladimirYuldashev\LaravelQueueRabbitMQ\Queue\RabbitMQQueue` to use this library.

I agree its not great that the constructors are no aligned but the fix is not a hack to make it work how you want it to but update to use factories to work with everyones code.

M-Porter:
Agreed with adm-bome.

Issue #527, opened 2023-02-23
Suggestion for a replacement to PhpAmqpLib\Connection\AMQPLazyConnection


maxacarvalho:
Hello,

I just started using the package and noticed that the class `PhpAmqpLib\Connection\AMQPLazyConnection` is flagged as deprecated. I wonder if there's any considerations about what implementation to use instead?
Thanks.

khepin:
Not at this time. The way to mark other connections as `lazy` is incompatible with how we create connections for now.

behzadsh:
So would you please update the README? The configuration in README is so confusing.

M-Porter:
@behzadsh you can always submit a PR to suggest fixes you think are needed.

PR #528, opened 2023-03-11
Fix/refactor creation of connections and queues

adm-bome:
Refactor:

Creating Conection with config via a factory. 
Creating Queue with config via a factory.

> Prepaired the library for deprication of the AMQPLazyConnection:class

#513, #521 and #527

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,9 +254,83 @@ class RabbitMQJob extends BaseJob
 }
 ```
 
+### SSL Secure
+
+When u need a secure connection to rabbitMQ server(s). Then this is posible by settitng extra config options.
+
+```php
+'connections' => [
+    // ...
+
+    'rabbitmq' => [
+        // ...
+
+        'secure' = > true,
+        'options' => [
+            // ...
+
+            'ssl_options' => [
+                'cafile' => env('RABBITMQ_SSL_CAFILE', null),
+                'local_cert' => env('RABBITMQ_SSL_LOCALCERT', null),
+                'local_key' => env('RABBITMQ_SSL_LOCALKEY', null),
+                'verify_peer' => env('RABBITMQ_SSL_VERIFY_PEER', true),
+                'passphrase' => env('RABBITMQ_SSL_PASSPHRASE', null),
+            ],
+        ],
+    ],
+
+    // ...    
+],
+```
+
+### Lazy Connection
+
+By default, your connection will be created as a lazy connection.
+If for some reason you dont want the connection lazy you can turn it off by setting the following config.
+
+```php
+'connections' => [
+    // ...
+
+    'rabbitmq' => [
+        // ...
+
+        'lazy' = > false,
+    ],
+
+    // ...    
+],
+```
+
+### Heartbeat
+
+By default, your connection will be created with a heartbeat setting of `0`.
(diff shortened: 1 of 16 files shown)


src/Queue/Connection/ConfigFactory.php
```diff
@@ -0,0 +1,93 @@
+<?php
+
+namespace VladimirYuldashev\LaravelQueueRabbitMQ\Queue\Connection;
+
+use Illuminate\Support\Arr;
+use PhpAmqpLib\Connection\AMQPConnectionConfig;
+
+final class ConfigFactory
```
khepin, line 8: This package used to make extensive use of `final` a couple years back and moved away from it. 
I'd prefer staying away from marking things `final`

khepin (commented):
I think the main structure is good.

The only thing I think should be updated is the use of `final` and `private`.

As it stands today this package uses the `private` keyword only once and never ever uses `final`. If memory serves well, a couple years ago, the opposite was true. I think it better to keep the same approach here.

README.md
```diff
@@ -254,17 +262,144 @@ class RabbitMQJob extends BaseJob
 }
 ```
 
+### Use your own Connection
+
+You can extend the build-in `PhpAmqpLib\Connection\AMQPStreamConnection::class`
```
mattgrande, line 267:
```suggestion
You can extend the built-in `PhpAmqpLib\Connection\AMQPStreamConnection::class`
```

README.md
```diff
@@ -254,17 +262,144 @@ class RabbitMQJob extends BaseJob
...
+        ],
+    ],
+
+    // ...    
+],
+```
+
+### SSL Secure
+
+When u need a secure connection to rabbitMQ server(s). Then this is possible by setting extra config options.
```
mattgrande, line 332:
```suggestion
If you need a secure connection to rabbitMQ server(s), you will need to add these extra config options.
```

mattgrande (approved):
Nice work on those factories; this is a huge improvement.

I've added a couple minor grammatical changes to the README, but other than that, this looks great.

M-Porter (approved):
This is great to see! I started working on a branch with factories but this is much better than any of the code I had.

Thank you for working on this!

README.md
```diff
@@ -254,17 +262,144 @@ class RabbitMQJob extends BaseJob
...
+
+    // ...    
+],
+```
+
 ## Laravel Usage
 
-Once you completed the configuration you can use the Laravel Queue API. If you used other queue drivers you do not need to
+Once you completed the configuration you can use the Laravel Queue API. If you used other queue drivers you do not need
+to
```
adm-bome, line 399:
```suggestion
Once you completed the configuration you can use the Laravel Queue API. If you used other queue drivers you do not need to
```

adm-bome:
@khepin @mattgrande @M-Porter

Could you guys, please have a new look at this PR.
I added some more minor changes / fixes.

If all is ok and **tested**, this can be merged. [version: v13.3.0]
@M-Porter can you make the neccesarry updates in the CHANGELOG when released?

khepin (approved):

khepin:
Thanks @adm-bome this is great

Issue #529, opened 2023-03-13
Can't add a delay between consumption of messages.


gssj85:
- Laravel version: 10.3.2
- RabbitMQ version: 3.5.2
- Package version: 13.2.0

Using Laravel's Mail Facade I sent 10 emails to the emails queue.

`$email = (new TaskEventsMail($message, $subject))->onConnection('rabbitmq')->onQueue('emails');
        for ($i = 0; $i < 10; $i++) {
            Mail::to($to)->send($email);
        }`

When I try to consume with:

`php artisan rabbitmq:consume --rest 5 --queue emails`

The emails are processed without the delay of 5 seconds between each.

adm-bome:
Because --rest option isn't implemented into the consumer `VladimirYuldashev\LaravelQueueRabbitMQ\Consumer:class`

This build in Consumer is made for high performance, so working with --rest is not the purpose of this worker.

**Tip 1:** 
>Try the laravel build in consumer on a rabbitmq connection and defined queue. See: ReadMe section Consuming messages.

**Tip2 :**
>Write your own consumer.

gssj85:
I think it's misleading then having the option there...

Anyway I'm kinda lost on how to create a consumer following this package Laravel itself standards, only worked with Symfony and used a bore bones amqp, any guidance would be appreciated.

khepin:
@adm-bome it's not implemented but I'm wondering if it should. The `rabbitmq:consume` command extends the `queue:work` one so it might be reasonable to implement all options from `queue:work` on `rabbitmq:consume`.
I'll re-open it for now so I can take a look later

adm-bome:
Sure we can implement it. No problem just 3 lines of code :wink: 

But (my opinion) adding a delay (sleep) to this worker, the purpose of this consumer will become some what blurry and you can just use the `queue:work` command. 

`basic_consume` does not work the same as `basic_get`. (not everything is possible)

khepin:
PR with my thoughts for why we should consider respecting it https://github.com/vyuldashev/laravel-queue-rabbitmq/pull/530

gssj85:
I got the delay I wanted this way:

1. Created a CustomMailJob class that extends SendQueueMailable
2. Overwrited the method handle() and before calling parent::handle() added a sleep(5)
3. In the Mailable class overwrited the method newQueueJob() and where it called SendQueuedMailable class, changed to the CustomMailJob class created.

[...]

PR #530, opened 2023-03-13
Honor the 'rest' option

khepin:
From https://github.com/vyuldashev/laravel-queue-rabbitmq/issues/529

Looks like we correctly honor all of the other `queue:work` options and have kept maintaining compatibility with `queue:work`.
There are also other options from `queue:work` that we honor despite the fact that they make little sense with this consumer. For example `sleep` when the queue is empty is useful for queues where you pull the work from the queue. Especially ones like SQS where you get charged every time you make a request. But when doing `basic_consume` this makes little sense yet it is respected by this package.

Our other choices with this `--rest` options would be:
1. accept it but ignore it (current situation). Making it the first of the `queue:work` options we are not compatible with
2. remove it (would break compatibility with `queue:work` by throwing an exception for providing an undefined option)
3. accept it and output a warning

### DIFF
--- a/src/Consumer.php
+++ b/src/Consumer.php
@@ -125,6 +125,10 @@ class Consumer extends Worker
                 if ($this->supportsAsyncSignals()) {
                     $this->resetTimeoutHandler();
                 }
+
+                if ($options->rest > 0) {
+                    sleep($options->rest);
+                }
             },
             null,
             $arguments
--- a/src/Consumer.php
+++ b/src/Consumer.php
@@ -127,7 +127,7 @@ class Consumer extends Worker
                 }
 
                 if ($options->rest > 0) {
-                    sleep($options->rest);
+                    $this->sleep($options->rest);
                 }
             },
             null,


adm-bome (approved):

src/Consumer.php
```diff
@@ -125,6 +125,10 @@ function (AMQPMessage $message) use ($connection, $options, $connectionName, $qu
                 if ($this->supportsAsyncSignals()) {
                     $this->resetTimeoutHandler();
                 }
+
+                if ($options->rest > 0) {
+                    sleep($options->rest);
```
adm-bome, line 130: Missing $this->  ?
>Some more magic in the $this->sleep()

adm-bome (changes requested):

src/Consumer.php
khepin: Absolutely! Thanks.

adm-bome (approved):

Issue #437, opened 2021-08-30
 local.ERROR: Trying to access array offset on value of type null 


farshidrezaei:
- Laravel/Lumen version: 8.57

I run `rabbit:consume rabbit --queue=test` , get this error in my logs:

> `[2021-08-30 11:53:22] local.ERROR: Trying to access array offset on value of type null {"exception":"[object] (ErrorException(code: 0): Trying to access array offset on value of type null at */vendor/laravel/framework/src/Illuminate/Queue/Jobs/Job.php:315)`

ijunaidali:
same error

sky93:
Same

sky93:
It's because your payload is an invalid Laravel payload schema and Laravel cannot parse it and it throws an error. If you use a custom payload not coming from another Laravel app, you have to extend the built-in class and create the payload yourself using this link: https://github.com/vyuldashev/laravel-queue-rabbitmq/blob/master/README.md#use-your-own-rabbitmqjob-class

angwa:
Please, I have this error too. But in my case, I already created a custom job class and extended rabbitmq job and added it to the queue config. 
`  'job' => \App\Jobs\RabbitMQJob::class,`

But I still face the same error.  Any suggestion?

geskeitaloricardo:
Just in case there is someone still getting this error, in my case the `getName` function was the problem!
Do like @angwa did and override getName function in \App\Jobs\RabbitMQJob::class:
`'job' => \App\Jobs\RabbitMQJob::class,`

```
use VladimirYuldashev\LaravelQueueRabbitMQ\Queue\Jobs\RabbitMQJob as BaseJob;

RabbitMQJob extends BaseJob {
  // override
  public function getName()
  {
       return "";
  }
}
```

then run `php artisan queue:work`

angwa:
Thank you for your response @geskeitaloricardo . After a lot of struggles. I noticed that Laravel has a specific format for receiving jobs. I was sending the job from a python script to Laravel as a plain text without formatting it to Laravel job. Thus, I was getting that error. I changed the message I was sending from python to contain job id and other stuff and got it working

M-Porter:
Closing as seeming to be resolved.

PR #532, opened 2023-03-21
[13.2] Add job example for handling non-json messages from rabbitmq queue

trckster:
I was having troubles trying to consume raw message like `Hello`. After long research I found closed issue #437. The error was the same and [comment](https://github.com/vyuldashev/laravel-queue-rabbitmq/issues/437#issuecomment-1228055468) helped me to fix problem. I think i'ts quite common case, to handle raw message, non JSON or without special Laravel fields. Therefore adding this to README might save developers time.


trckster:
Resolved merge conflicts. What do you think about this documentation upgrade?

mattgrande (approved):
Thank you for this!

PR #554, opened 2023-07-07
Fix #553, add possibility to define network protocol in config

### LINKED ISSUE #553: Network Protocol issue with 13.3.X version
- Laravel version: 9
- Package version: 13.3.X

**Describe the bug**

In previous version of this package ( < 13.3.X ), the default network protocol used when using `AMQPSSLConnection` was `ssl`.
Now with the current implementation of the package, it use the default value of `AMQPConnectionConfig` which is `tcp`.

**Steps To Reproduce**

Use default connection with `'secure' => true`.

**Current behavior**

On our configuration, we receive the error `Invalid frame type 21`

**Expected behavior**

We should be able to configure the used network protocol.

**Pull Request**

I created a PR for this case

maximetassy:
close #553

### DIFF
--- a/README.md
+++ b/README.md
@@ -513,6 +513,7 @@ If for some reason you don't want the connection lazy you can turn it off by set
 
 By default, the network protocol used for connection is tcp.
 If for some reason you want to use another network protocol, you can add the extra value in your config options.
+Available protocols : `tcp`, `ssl`, `tls`
 
 ```php
 'connections' => [
@@ -521,11 +522,7 @@ If for some reason you want to use another network protocol, you can add the ext
     'rabbitmq' => [
         // ...
 
-        'options' => [
-            // ...
-
-            'network_protocol' => 'network wanted',
-        ],
+        'network_protocol' => 'tcp',
     ],
 
     // ...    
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -509,6 +509,29 @@ If for some reason you don't want the connection lazy you can turn it off by set
...
+'connections' => [
+    // ...
+
+    'rabbitmq' => [
+        // ...
+
+        'options' => [
+            // ...
+
+            'network_protocol' => 'network wanted',
```
adm-bome, line 527: I rather see this one level up. And accessed via

```php
Arr::get($config, 'network_protocol')
```

Because network connection config is defined next to `lazy` and `secure` config parameters.

The `options` are more intended for config in the worker Class.
> Now this is a bit misleading, as whe have to handle legacy config structure previous to v13.3.0

adm-bome (changes requested):

README.md
adm-bome: And maybe document the available protocols

maximetassy: Thanks for your review, I updated the PR with your suggestions

adm-bome (approved):

PR #560, opened 2023-07-18
Update composer.json

adm-bome:
The minimal required version for methods: getSslCaPath(), getSslSecurityLevel() in the connectionConfig.


adm-bome:
Master was updated by other PR

PR #564, opened 2023-07-27
Create delayed queues as quorum when it's enabled

### LINKED ISSUE #563: Delayed queues as type quorum queue
First, let me say thank you for this great package!

We're using Delayed queues, the queues are created by this package and also disappear when they are not needed anymore, this is all working fine!

Some of our queues, will delay messages for a couple of days, for example: When an external service is down.
I saw that that created `default.delay.xyz` queues, are of the type `classic`.

Is it Possible to use the configuration option if quorum enabled?
It's a really small change in the `getDelayQueueArguments`, so I've created a PR.

TomKeur:
Fixes #563

### DIFF
--- a/src/Queue/RabbitMQQueue.php
+++ b/src/Queue/RabbitMQQueue.php
@@ -620,7 +620,6 @@ class RabbitMQQueue extends Queue implements QueueContract, RabbitMQQueueContrac
      */
     protected function getDelayQueueArguments(string $destination, int $ttl): array
     {
-
         $arguments = [
             'x-dead-letter-exchange' => $this->getExchange(),
             'x-dead-letter-routing-key' => $this->getRoutingKey($destination),
--- a/src/Queue/RabbitMQQueue.php
+++ b/src/Queue/RabbitMQQueue.php
@@ -620,12 +620,19 @@ class RabbitMQQueue extends Queue implements QueueContract, RabbitMQQueueContrac
      */
     protected function getDelayQueueArguments(string $destination, int $ttl): array
     {
-        return [
+
+        $arguments = [
             'x-dead-letter-exchange' => $this->getExchange(),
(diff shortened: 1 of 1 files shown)


khepin:
I've enabled the test runs. Can you take a look at the failures there.

TomKeur:
Thanks for the quick reply! I've fixed the styling issue, somehow the last version of Laravel Pint did not come up this error, but I've fixed it by using the same composer command used in the GitHub Action workflow :).

khepin (approved):

khepin:
@adm-bome any thoughts? If not I'll merge tomorrow.

TomKeur:
For your info: today I've upgraded to RabbitMQ 3.12.2, now I'm able to set the Default Queue type per vhost, when it's set to quorum, every queue generated will be a quorum queue.

But I guess it's still a nice addition for people working with an older version of RabbitMQ.

adm-bome:
@khepin I will look in to it. (Will comment in 2 à 3 days)

TomKeur:
Any news?

adm-bome:
My concern.

When enabling delay queues, an inline option could be set to also make delay queues quorum. When you want to and quorum was set for the processing queue.

Also testing for this config and the functional part would be nice addition.

TomKeur:
After doing some research, it seems like for the default we do not want to use Quorum queues for Delay queues, see:

https://www.rabbitmq.com/quorum-queues.html#use-cases

"When Not to Use Quorum Queues", 

> Temporary nature of queues: transient or exclusive queues, high queue churn (declaration and deletion rates)

But since RabbitMQ 3.12 you can set the default on vhost level in the GUI, when creating a new vhost.
So maybe it should be an option, to set the "Default" for delay queues, for example

`'x-queue-type' => 'classic',`

adm-bome:
So, it makes no sense and will not add any performance to the processing of messages.

PR #604, opened 2024-10-22
Added requeue parameter for markAsFailed()

urlportal:
`reject(RabbitMQJob $job, bool $requeue = false)` supports second parameter $requeue. I suggest adding a second parameter to make the method `markAsFailed()` more flexible.


khepin:
This would be incompatible with the framework implementation: https://github.com/laravel/framework/blob/11.x/src/Illuminate/Queue/Jobs/Job.php#L171

PR #609, opened 2024-11-26
Use config option for declare exchange and queue

imanabbasi:


khepin:
Can you update so the tests pass?

src/Queue/RabbitMQQueue.php
```diff
@@ -717,7 +721,11 @@ protected function declareDestination(string $destination, string $exchange = nu
         }
 
         // Create a queue for amq.direct publishing.
-        $this->declareQueue($destination, true, false, $this->getQueueArguments($destination));
+        $options = $this->config->getOptions();
+        $durable = Arr::get($options, 'durable') ?: true;
+        $autoDelete = Arr::get($options, 'auto_delete') ?: false;
```
FrittenKeeZ: Why not use 3rd argument of `Arr::get()`?
```php
$durable = Arr::get($options, 'durable', true);
$autoDelete = Arr::get($options, 'auto_delete', false);
```

PR #623, opened 2025-04-21
Fix PHP 8.4 deprecated nullable parameter warnings

softwaredeveloper97:
This PR updates the method signatures to use explicit nullable types in accordance with PHP 8.4 deprecation notices.


khepin (approved):

mattgrande (approved):

khepin:
It's failing on the `pint` step to use the predefined styles. Let's update that.

fontebasso:
Thanks, @softwaredeveloper97. This has already been addressed in PR #616, so you can go ahead and close this one.

Issue #601, opened 2024-09-12
Quorum queues for delayed queues to ensure *at-least-once* delivery


philipheimboeck:
* This change requires at least RabbitMQ 3.10.
* Version used: 14.1

Hello everyone!

In a clustered environment it can happen that messages are being lost when using dead-letter-exchanges. Now this also means that when using delayed messages, it can happen that messages would get lost as the documentation states:

> By default, dead-lettered messages are re-published without publisher [confirms](https://www.rabbitmq.com/docs/confirms) turned on internally. Therefore using DLX in a clustered RabbitMQ environment is not guaranteed to be safe.

https://www.rabbitmq.com/docs/dlx#safety

With quorum queues it is possible to enable `at-least-once` dead-lettering for a source quorum queue.

https://www.rabbitmq.com/docs/quorum-queues#dead-lettering

I suggest a change, where when the `quorum` option is enabled, the dead letter queues are also created as quorum queus and additionally enable the settings to for at-least-once delivery. The motivation behind quorum queues is to focus on data safety, so I think the same should apply to delayed queues.

Please note that this topic was also already discussed in #564, but it was closed due to not bringing any benefits. However I think the reasoning was a different one as it was about performance and not data safety.

I have tested the following patch locally with our system. Please note that I also disabled the `x-expires` option, which means the queues won't disappear anymore.

This is due to two reasons:
1. Quorum queues are not really for short-lived queues, therefore it makes more sense to keep them alive. Also in many use cases the delays will be the same many times, so there shouldn't be too many new queues. Of course this is a topic which has to be discussed and heavily depends on the use cases, but it might be still be an ok limitation if quorum queues are needed.
2. In my tests I pushed many messages to the queues and it happened that the client crashed, because a message was pushed while it was removed (tested with RabbitMQ 3.13.7). By removing this that problem was gone.

```patch
--- /dev/null
+++ ../src/Queue/RabbitMQQueue.php
@@ -625,11 +625,21 @@
      */
     protected function getDelayQueueArguments(string $destination, int $ttl): array
     {
+        $arguments = [];
+        $isQuorum = $this->getConfig()->isQuorum();
+        if ($isQuorum) {
+            $arguments = [
+                'x-queue-type' => 'quorum',
+                'x-dead-letter-strategy' => 'at-least-once',
+                'x-overflow' => 'reject-publish',
+            ];
+        }
         return [
+           ...$arguments,
[...]

PR #627, opened 2025-04-29
Added config option to allow disabling x-expires in `getDelayQueueArguments`

cosminardeleanu:
I am using quorum and i noticed that the messages are deleted before they're moved to their destination queue.
By disabling the `x-expires`, my messages are no longer lost.

I think is considerate to have the ability to enable/disable delay queue expires, to avoid issues for users that use quorum.

Found other issue related -> https://github.com/vyuldashev/laravel-queue-rabbitmq/issues/601

----

### From the TTL documentation (/ttl):
This page describes the key difference for x-expires. 

> "Queues will expire after a period of time only when they are not used (a queue is used if it has online consumers). ... This controls for how long a queue can be unused before it is automatically deleted."

### From the Quorum Queues documentation (/quorum-queues):
This page describes the core design that explains why the TTL behavior is different.

> "Quorum queues are optimized for certain use cases where data safety is the top priority."
"Quorum queues have differences in behaviour compared to classic queues as well as some limitations that it is important to be aware of..."

When combined, these two pages explain that the "unused" criteria for Classic queues (which prevents deletion if messages are present) does not apply to the more robust, data-safe Quorum queues, which prioritize consistency and guaranteed cleanup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,30 @@ by adding extra options.
 ],
 ```
 
+When you want to disable x-expires from a delayed queue, then this is possible by adding extra option on queue `expire_delay_queue`.
+- When the `expire_delay_queue` option is omitted, it is considered to be true.
+- Useful for when you're using quorum queues, as they risk to delete the queue before the messages are moved to their destination queue.
+
+```php
+'connections' => [
+    // ...
+
+    'rabbitmq' => [
+        // ...
+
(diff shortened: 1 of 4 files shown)


makkmarci13:
Any update about this issue? Would be really useful if it will be available soon.

Issue #630, opened 2025-05-12
Strange RabbitMQ errors


akalongman:
Versions: 
- Laravel/Lumen version: 12.13.0
- RabbitMQ version: 3.9.21
- Package version: 14.2.0

I am getting strange errors and I dont know how to fix them.  Even how to debug.

I have ran queue with supervisord:
```yaml
[program:project-queue]
process_name=%(program_name)s_%(process_num)02d
command=php /var/www/project.ge/artisan queue:work --queue=queue1,queue2,queue3 --tries=3 --max-time=3600
autostart=true
autorestart=true
stopasgroup=true
killasgroup=true
user=project
numprocs=16
redirect_stderr=true
stdout_logfile=/var/www/project.ge/storage/logs/workers/queue.log
stopwaitsecs=3600
```

And in Sentry I have the following errors:

1. `PhpAmqpLib\Exception\AMQPRuntimeException`: Lost connection: Channel connection is closed.
2. `PhpAmqpLib\Exception\AMQPProtocolChannelException`: PRECONDITION_FAILED - delivery acknowledgement on channel 1 timed out. Timeout value used: 1800000 ms. 
3. `PhpAmqpLib\Exception\AMQPTimeoutException`: The connection timed out after 3 sec while awaiting incoming data.

Queue Config:
```php
'rabbitmq' => [
    'driver' => 'rabbitmq',
    'queue'  => env('RABBITMQ_QUEUE', 'default'),
    'hosts' => [
        [
            'host'     => env('RABBITMQ_HOST', '127.0.0.1'),
            'port'     => env('RABBITMQ_PORT', 5672),
            'user'     => env('RABBITMQ_USER', 'guest'),
            'password' => env('RABBITMQ_PASSWORD', 'guest'),
            'vhost'    => env('RABBITMQ_VHOST', '/'),
        ],
    ],
    'options' => [
        'queue' => [
            'job' => RabbitMQJob::class,
        ],
    ],
    'connection'   => RabbitMQConnection::class,
    /*
     * Set to "horizon" if you wish to use Laravel Horizon.
     */
    'worker'       => RabbitMQQueue::class,
    'after_commit' => true,
],
```

The content of the `RabbitMQQueue`:

```php
<?php

declare(strict_types=1);

namespace App\Libraries\Queue;

use PhpAmqpLib\Channel\AMQPChannel;
use Throwable;
[...]

PR #612, opened 2024-12-12
load timeouts from laravel config

Triplkrypl:
https://github.com/vyuldashev/laravel-queue-rabbitmq/issues/586

Issue #598, opened 2024-08-09
Keepalive not being set?


TomaszKorwel:
I have keepalive set to true in queue.php yet the ConfigFactory seems to be ignoring that setting and not passing that on to Connection.

Is there anything I need to set differently to have that setting passed?

Issue #596, opened 2024-07-14
Lazy connections doesn't work


Excelautomation:
- Laravel/Lumen version: 10.48.16
- RabbitMQ version: 3.12.12
- Package version: 13..3.5

My issue is that all queue workers are running and keeping the connections open. Right now I have 300 connections open, and CloudAMQP is beginning to sent notifications about too many open connections. 

In the docs, you write that the connections are default lazy connections. 

How can I make sure that the connections are only open, if they are actually have jobs to process?

I have this queue.php:
```
        'rabbitmq' => [
            'driver' => 'rabbitmq',
            'worker' => \App\Queue\RabbitMQQueue::class,
            'connection' => PhpAmqpLib\Connection\AMQPStreamConnection::class,
            'lazy' => true,
            'hosts'  => [
                [
                    'host'     => env('RABBITMQ_HOST', '127.0.0.1'),
                    'port'     => env('RABBITMQ_PORT', 5672),
                    'user'     => env('RABBITMQ_USER', 'guest'),
                    'password' => env('RABBITMQ_PASSWORD', 'guest'),
                    'vhost'    => env('RABBITMQ_VHOST', '/'),
                ],
            ],
        ],
```

And this RabbitMQQueue.php:
```
<?php

namespace App\Queue;

use PhpAmqpLib\Channel\AMQPChannel;
use PhpAmqpLib\Exception\AMQPChannelClosedException;
use PhpAmqpLib\Exception\AMQPConnectionClosedException;
use VladimirYuldashev\LaravelQueueRabbitMQ\Queue\RabbitMQQueue as BaseRabbitMQQueue;

class RabbitMQQueue extends BaseRabbitMQQueue
{

    protected function publishBasic($msg, $exchange = '', $destination = '', $mandatory = false, $immediate = false, $ticket = null): void
    {
        try {
            parent::publishBasic($msg, $exchange, $destination, $mandatory, $immediate, $ticket);
        } catch (AMQPConnectionClosedException|AMQPChannelClosedException) {
            $this->reconnect();
            parent::publishBasic($msg, $exchange, $destination, $mandatory, $immediate, $ticket);
        }
    }

    protected function publishBatch($jobs, $data = '', $queue = null): void
    {
        try {
            parent::publishBatch($jobs, $data, $queue);
        } catch (AMQPConnectionClosedException|AMQPChannelClosedException) {
            $this->reconnect();
            parent::publishBatch($jobs, $data, $queue);
[...]

PR #632, opened 2025-06-08
Add blocking mode with auto-reconnection and some other features.

Xfaider48:
I'm glad to introduce a lot of features that were made in the driver working on other projects.

Here is the short list:
1. The `blocking` mode option [Consumer]
2. The auto-reconnection option [Consumer]
3. The initializing queue option [Consumer]
4. The alive-check option [Consumer]
5. More debug/error messages [Consumer]
6. The verbose-messages option [Consumer]
7. Retry options [Queue]
8. Another expiration strategy [Queue]
9. Support `ttl` less than a secong [Queue]
10. Option to cache declared [Queue]
11. Option to declare full route [Queue]

Some of the features were added to improve a perfomance of hanling messages using Consumer, and the others were added to improve durability.

There options are production ready (I migrated them from the v13, so there could be some bugs becasue of that).
These changes don't break anything and disabled by default (except the point 5 about new messages)
Now let me explain why here a lof ot options added.

### Increasing perfomance 
In the default mode (non-blocking) when you don't have any messages in a queue the consumer tries to get a job, and if gets nothing it sleeps, but while it sleeps we can get a job in a queue and that job will wait, so we lose a lot of perfomance and the consumer consumes resources like CPU. To fix the problem **the blocking** mode was added.

When you use blocking mode the worker just waits for some time (or forever) for new messages, and while it waits it consumes no resources and when a job comes the worker gets it ASAP and starts working on it.

And the last one about perfomance: in the production we don't need any simple logs about successfuly processed jobs, but about errors. To fix it the **verbose-messages option** was added, you can use it to write messages ONLY if you start worker in the verbose mode (`-v`)

### Increasing durability
We don't live in the perfect world, shit happens sometime and we need to try to handle as much as we can.

To increase the Consumber durability we need to enable the **auto-reconnection option**. 
When something is wrong with the connection the consumer will automatically reconnect. Also we can use `initializing queue option`, because if the problem happens because of the RabbitMQ and the RabbitMQ restarted (you might don't have anything in it) you will get an error after reconnecting becasue the queus doesn't exist anymore.

When you use the **blocking mode** the connection can stuck in some strange state (the consumer works, the RabbitMQ sees the consumer, but messages don't reach the consumer at all), it could happen when you use a complex architecture with vurtial machines and/or proxies. To check that the connection is stuck we must call some action and wait for the response, for this the **alive-check option** was added. Just set N seconds and the consumer will check that all is OK with the connection every N seconds of working (it happens when NO messages comes in N seconds, so this doesn't affect the perfomance a lot).
Also the `keepalive` option must be enabled for the blocking mode, that's why it's added to the options.

And the last thing for the consumer: I added more logs in it to simplify an investigation about what happened, because sometimes the consumer dies and you don't get anything about what was going on (maybe it was stopped because of some options, maybe it was killed by the system)

## Improving the Queue
First of all I want to speak about the **option to cache declared**. The origial code doesn't have it and it caches all declared exchanges and queues in the memory, but it's completely WRONG (see the section about **auto-reconnection option** above), because the original code doesn't handle any problems with the connection and don't clear them in the memory. Now it's fixed and the option was added just in case to simplify any investigations of some problems of caching (useful with the Laravel Octane).

The **retry options** was added to automatically reconnect and try again if we get some problems while we send messages, it helps to avoid problems during short temporary network problems. (Maybe it's better to make another class for it, but now it's included in the Queue class)

The **another expiration strategy** was added to impove latering messages. The original strategy creates a lot of queues (one queue per TTL), but the new one just uses the [expiration](https://www.rabbitmq.com/docs/ttl#per-message-ttl-in-publishers) option of messages, and the latering messages will become more effective with this option enabled.

The **option to declare full route** was added to improve durability and stability. If you face some problems with connection and the RabbitMQ lost everything you might want to restore the full route of message, right? So, you can enable this option and before first sending the Queue class will declare exchange + queue + bind them, and you will get all things work.

## What else can be improved
- I think we need to refactor README and describe all possible options.
- We need to add more options for connections and handle all of them (see https://github.com/php-amqplib/php-amqplib/blob/master/PhpAmqpLib/Connection/AMQPConnectionConfig.php)

## Affected PRs
- https://github.com/vyuldashev/laravel-queue-rabbitmq/pull/414 can be closed becase of **blocking mode** added
- https://github.com/vyuldashev/laravel-queue-rabbitmq/pull/612 must be fixed accorging to added RPC timeout
- https://github.com/vyuldashev/laravel-queue-rabbitmq/pull/627  must be fixed accorging to the new expiration strategy added

## Affected issues
- https://github.com/vyuldashev/laravel-queue-rabbitmq/issues/630 might be affected and solved because of **auto-retries** added in the core
- https://github.com/vyuldashev/laravel-queue-rabbitmq/issues/596 might be affected and solved because of **auto-retries** added in the core
- https://github.com/vyuldashev/laravel-queue-rabbitmq/issues/598 will be resolved

### DIFF
--- a/src/Consumer.php
+++ b/src/Consumer.php
@@ -3,12 +3,20 @@
 namespace VladimirYuldashev\LaravelQueueRabbitMQ;
 
 use Exception;
+use Illuminate\Console\Concerns\InteractsWithIO;
 use Illuminate\Container\Container;
+use Illuminate\Contracts\Debug\ExceptionHandler;
+use Illuminate\Contracts\Events\Dispatcher;
+use Illuminate\Contracts\Queue\Factory as QueueManager;
+use Illuminate\Queue\Events\JobTimedOut;
 use Illuminate\Queue\Worker;
 use Illuminate\Queue\WorkerOptions;
 use PhpAmqpLib\Channel\AMQPChannel;
-use PhpAmqpLib\Exception\AMQPRuntimeException;
+use PhpAmqpLib\Exception\AMQPExceptionInterface;
+use PhpAmqpLib\Exception\AMQPProtocolChannelException;
+use PhpAmqpLib\Exception\AMQPTimeoutException;
 use PhpAmqpLib\Message\AMQPMessage;
+use Symfony\Component\Console\Output\OutputInterface;
 use Throwable;
 use VladimirYuldashev\LaravelQueueRabbitMQ\Queue\RabbitMQQueue;
 
@@ -29,11 +37,69 @@ class Consumer extends Worker
     /** @var int */
     protected $prefetchCount;
 
-    /** @var AMQPChannel */
+    /**
+     * @var bool
+     */
+    protected $blocking = false;
+
+    /**
+     * @var bool
+     */
+    protected $initQueue = false;
+
+    /**
+     * @var bool
+     */
+    protected $autoReconnect = false;
+
+    /**
+     * @var float
+     */
+    protected $autoReconnectPause = 0;
+
+    /**
+     * @var float
+     */
+    protected $aliveCheck = 0;
+
+    /**
+     * @var bool
+     */
+    protected $asyncSignalsSupported;
+
+    /**
+     * @var RabbitMQQueue
+     */
+    protected $connection = null;
+
+    /**
+     * @var AMQPChannel
+     */
     protected $channel;
 
-    /** @var object|null */
+    /**
+     * @var object|null
+     */
     protected $currentJob;
+
+    /**
+     * @var InteractsWithIO
+     */
+    protected $interactsWithIO = null;
+
+    /**
+     * {@inheritDoc}
+     */
+    public function __construct(
+        QueueManager $manager,
+        Dispatcher $events,
+        ExceptionHandler $exceptions,
+        callable $isDownForMaintenance,
+        ?callable $resetScope = null
+    ) {
+        parent::__construct($manager, $events, $exceptions, $isDownForMaintenance, $resetScope);
+        $this->asyncSignalsSupported = $this->supportsAsyncSignals();
+    }
 
     public function setContainer(Container $value): void
     {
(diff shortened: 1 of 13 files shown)


Xfaider48:
@iamfarhad If there is no any activities from maintainers I can adapt this for your version

Xfaider48:
@khepin Could you please review this?

FrittenKeeZ:
@Xfaider48 is this tested with Horizon as well?

FrittenKeeZ:
@Xfaider48 why I'm asking regarding Horizon is because we're encountering a large amount of `PhpAmqpLib\Exception\AMQPConnectionClosedException` exceptions and wanted to know if this PR resolves the same issues, or if I should implement the reconnect example from the docs (https://github.com/vyuldashev/laravel-queue-rabbitmq?tab=readme-ov-file#use-your-own-worker-class).

src/Consumer.php
```diff
@@ -193,18 +415,226 @@ protected function daemonShouldRun(WorkerOptions $options, $connectionName, $que
...
+            // We need to call any command that has a response, and we can wait for it.
+            // If there is no response, we consider the current connection dead
+            // WARNING: It works ONLY if `channel_rpc_timeout` is greater than 0
+            $this->channel->basic_qos(
+                $this->prefetchSize,
+                $this->prefetchCount,
+                null
+            );
+        } catch (AMQPTimeoutException $exception) {
+            throw new AMQPTimeoutException('Custom alive check failed', $exception->getTimeout(), $exception->getCode(), $exception);
```
FrittenKeeZ: Is there any reason to catch this just to throw it again?

Xfaider48: To set up the message correctly

FrittenKeeZ (approved):

FrittenKeeZ:
@khepin this would be a great first release for the new year ;)

PR #641, opened 2025-11-18
Change getQueueArguments to public and allow custom failed exchange to be passed

FrittenKeeZ:
When declaring custom queues it's nice to be able to use the existing queue arguments method for handling failed jobs.


FrittenKeeZ:
@khepin will you have a look at this again?

PR #650, opened 2026-02-11
Add handling for setting connection name and make config publishable

FrittenKeeZ:
- Allows for static and dynamic setting of connection name, using either a `string` or a `Closure` which will receive the config array
- Allows for publishing the `rabbitmq.php` config and properly resolve merging if so


FrittenKeeZ:
@khepin when you have time it would be nice to hear your thoughts on this.

PR #669, opened 2026-05-27
 Fix: inherit WorkCommand signature to stay compatible with new Laravel options

### LINKED ISSUE #668: `The "stop-when-empty-for" option does not exist` error on Laravel 13.10.+
- Laravel/Lumen version: 13.11.2
  - RabbitMQ version: N/A (failure occurs before any broker connection is made)
  - Package version: 15.0.0

  **Describe the bug**
  
  `VladimirYuldashev\LaravelQueueRabbitMQ\Console\ConsumeCommand` extends `Illuminate\Queue\Console\WorkCommand` and re-declares its own `$signature` to add RabbitMQ-specific options (`--max-priority`, `--consumer-tag`, `--prefetch-size`, `--prefetch-count`). However, the re-declared
  signature does **not** mirror the full list of options from `WorkCommand`.

  Starting with `laravel/framework` 13.10, `WorkCommand` introduced a new option, `--stop-when-empty-for=0`, and `WorkCommand::handle()` now unconditionally reads it via `$this->option('stop-when-empty-for')` (see `vendor/laravel/framework/src/Illuminate/Queue/Console/WorkCommand.php:173`).

  Because `ConsumeCommand` overrides `$signature` without including this new option, invoking `rabbitmq:consume` aborts with:

  ```
  The "stop-when-empty-for" option does not exist.
  ```
  
  In other words, the `ConsumeCommand` signature has drifted out of sync with its parent in `laravel/framework` >= 13.10.

  **Steps To Reproduce**
  
  1. Create or upgrade a Laravel application to `laravel/framework: ^13.10` (we hit it on `v13.11.2`).
  2. Install `vladimir-yuldashev/laravel-queue-rabbitmq: ^15.0` (`v15.0.0`).
  3. Configure a `rabbitmq` queue connection per the package README.
  4. Run any worker command, e.g.:

     ```bash
     php artisan rabbitmq:consume rabbitmq --queue=default
     ```

  5. The command immediately fails with `The "stop-when-empty-for" option does not exist.` — it never connects to RabbitMQ.

  **Current behavior**
  
  - The worker process exits immediately on startup with a non-zero exit code; no jobs are processed.
  - No connection to RabbitMQ is opened, so nothing happens to existing messages on the queue — they remain in their current state (Ready / Unacked / etc.) untouched.
  - Because the failure is at command-parsing time (before `WorkCommand::handle()` runs), no message is consumed, acknowledged, rejected, or requeued by this worker invocation.
  - Under a process supervisor (Supervisor / systemd / Kubernetes), this manifests as a crash-loop: the worker keeps restarting and exiting with the same error.

  **Expected behavior**

  `rabbitmq:consume` should start cleanly on `laravel/framework` >= 13.10 and accept the same `--stop-when-empty-for=` option as `queue:work` (default `0`, i.e. disabled), forwarding it to the worker so that "stop when no jobs have been processed for N seconds" behaves the same way as for
  the stock queue driver.

  Concretely, `ConsumeCommand::$signature` should include:

  ```
  {--stop-when-empty-for=0 : Stop when no jobs have been processed for the given number of seconds}
  ```
  
  (matching the line added to `WorkCommand` in Laravel 13.10).

  **Additional context**
  
  - Root cause is structural: `ConsumeCommand` redeclares `$signature` instead of extending the parent's signature, so every new option added upstream in `WorkCommand` silently breaks `rabbitmq:consume` until the package is updated. This is the second time this kind of drift has happened
  (see also previous additions like `--rest`, `--max-time`, `--json`).
  - A more durable fix would be to either (a) build the signature dynamically from `parent::$signature` plus the RabbitMQ-specific options, or (b) add a CI matrix job against the latest `laravel/framework` minor to catch new options as they land.
  - Workaround for affected users until a fix is released: add the missing option in a local subclass of `ConsumeCommand` and register it in place of the package's command.
  - Affects: any deployment running `laravel/framework` >= 13.10 with `vladimir-yuldashev/laravel-queue-rabbitmq` <= 15.0.0. The same defect likely affects older majors of the package that ship a `ConsumeCommand` predating Laravel 13.10.
  - Option added in https://github.com/illuminate/queue/commit/53f093270595f1c7828816f7e3174cc0848c370d
  -

vincedom:
Problem

  ConsumeCommand redeclared the entire WorkCommand signature, so it silently dropped any option Laravel added upstream. The recently introduced --stop-when-empty-for option makes
  WorkCommand::handle() read an option our signature never declared, so the worker crashes on startup with:

  The "stop-when-empty-for" option does not exist.

  This happens before any RabbitMQ connection is opened, so under a process supervisor (Supervisor, systemd, Kubernetes) it turns into a crash loop.

  I'm hitting this on Laravel 12.60.0 as well as 13.11.0.

  Fixes #668.

  Fix

  Rather than copying the parent signature (which guarantees recurring drift every time Laravel adds an option), this PR:

  - Inherits WorkCommand's signature instead of redeclaring it, so we automatically pick up current and future upstream options.
  - Renames the command to rabbitmq:consume and appends only the RabbitMQ-specific options (--max-priority, --consumer-tag, --prefetch-size, --prefetch-count).
  - Skips any option whose name the parent already defines, so a future upstream collision defers to Laravel instead of crashing on a duplicate definition.

  The RabbitMQ name and options are kept as readable DSL text in class constants and parsed via Illuminate\Console\Parser.


nargil07:
I tried this and work like a charm in our case :smile: 
Using it on Laravel 12.61

clemblanco:
Hey @khepin @vyuldashev can you guys please review it? 🙏

mcdado:
While waiting for an upgrade, we asked Fable 5 to patch our app.

SCR-20260612-kbjo

vincedom:
I have created a temporary package with this fix https://packagist.org/packages/vincedom/laravel-queue-rabbitmq  https://github.com/vincedom/laravel-queue-rabbitmq

clemblanco:
How long does it take for a maintainer to look at this usually?

vyuldashev:
Released under v15.0.1

